PR #2, opened 2016-01-07
Use all yml files in global, site, and environment level directories - not just the ones initially created

jrbudnack:
This allows me to use Genesis in existing repos, which may have other yml files that contain configuration options.  Genesis will still use the same structure when it creates a new project, but this allows the project developer to split things out into other files if needed and still have the project build.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -287,16 +287,18 @@ Global Definitions
 
 This directory contains templates that describe the common elements of this
 deployments, to be shared (and possibly overridden) by sites and environments.
+The templates are merged in the following order:
+
+  global/jobs.yml             Specify what jobs will make up the canonical
+                              deployment, and what templates to apply to them.
 
   global/deployment.yml       Define the global structure of all deployments
                               with the correct param calls to remind template
                               writers to override the correct things.
 
-  global/jobs.yml             Specify what jobs will make up the canonical
-                              deployment, and what templates to apply to them.
-
   global/properties.yml       Define the properties (global or per-job) for this
                               deployment.
+
 
 NOTE: If you make changes to the templates in here, they will automatically
 propagate to any newly-created environments, but you will need to run a
@@ -314,6 +316,7 @@ deployments, to be shared (and possibly overridden) by sites and environments.
                               deployments, with the correct param calls to remind
                               template writers to override the correct things.
 
+
 NOTE: If you make changes to the templates in here, they will automatically
 propagate to any newly-created environments, but you will need to run a
 refresh for existing environments to receive those updates.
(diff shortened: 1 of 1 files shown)


filefrog:
Awesome.

A few things:

1. Can you do this for the other types of deployments as well (microbosh and bosh-init)?
2. Can you sort the globbed YAML files so that the result is stable across all platforms (no idea of POSIX requires globs to present uniformly, or allows the fs layer to do what's most efficient) - I don't want to run this on my mac and have file a.yml override b.yml, only to find b.yml overriding a.yml on a Linux node...
3. Can you update the READMEs (for all three deployment types, global / site / environment layers) to point out that you can add additional files?

geofffranks:
Merge order is important for spruce.. are we sure we want to have the generated output potentially change, if someone renames a file?

filefrog:
AHA!  That was the reason we stayed away from globs.

@jrbudnack can we instead just add the site-level properties.yml file?

@geoff: how do you feel about creating empty files on `genesis build` if they don't exist?

jrbudnack:
@filefrog I can certainly do that. What I was trying to do is have the ability to add that properties.yml, but without messing up users whose repos that did not have that new file when they upgraded their own copy of Genesis and suddenly had "genesis build" fail.  I like your idea of creating an empty file if it does not exist - it seems to solve that problem.

geofffranks:
What about a default merge order/file listing (keep the auto-create empty files), but allow users to override the merge order via some sort of in-repo dot-file configuration?

Just a thought, not sure if i like it.

jrbudnack:
@geofffranks it may be a good format to adopt, not just for the dot-file paradigm, but for possibly adding more templates to Genesis as it evolves.

Along that vein, imagine the following directory structure for Genesis:

~/.genesis - A clone of the Genesis repo

~/bin/genesis -> ~/.genesis/bin/genesis (or wherever you want to symlink to the executable)

~/.genesis/templates:
   bosh-init
   - .build-order (or some better descriptive name), with the default entries
   - Any other template files
   microbosh
   (same as above, only for the microbosh flavor)
   normal_site
   (same as above, only for the normal_site flavor)

This would take much of the file content out of the code of the main executable, and possibly make Genesis easier to expand with additional project types.

jrbudnack:
Plus you could add "genesis upgrade" as a simple "git pull"

filefrog:
@jrbudnack you are, of course, referring to `genesis embed`

norman-abramovitz:
My initial concern was how to avoid files that should not be included if you used the globing.   The file list solves this problem as well

jrbudnack:
I'll put in a different pull request with the new site-level properties.yml file, along with logic that stubs out any missing yml files.

PR #3, opened 2016-01-08
Add properties.yml to site-level.  Add stub files if they don't exist.  Order README generation according to file merge order.

jrbudnack:
This will fix the problem for now so we can start using Genesis in our particular projects - but I do realize a flaw in this solution:  The files missing in .site and .global will be filled in with the missing stubs, but it won't do anything about adding stubs to ../site and ../../global for permanent inclusion.  I'm not sure how to best implement without maintaining multiple file lists within the code, making it harder to keep track of where to make alterations when needed.  Perhaps we can chat about it tomorrow.

For now, issuing the pull request.  We can tweak as we figure things out.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -883,7 +883,7 @@ update:
   update_watch_time: 1000-60000
   max_in_flight: 1
 EOF
-	for file in networks resource-pools jobs; do
+	for file in networks resource-pools jobs properties; do
 		cat > ${DEPLOYMENT_ROOT}/${site}/site/${file}.yml <<EOF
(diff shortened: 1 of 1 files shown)


filefrog:
Does it hurt us to just ignore files that are missing?  I.e. don't echo them from `ensure_yaml_files` unless the exist?

jrbudnack:
Changed per our discussion.  Files are now simply skipped if they do not exist.

filefrog:
Beauty! :+1:

PR #10, opened 2016-01-22
generalize sha1sum function so it will work on MacOS

johnlonganecker:
generalize sha1sum function so it will work on MacOS

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1291,6 +1291,14 @@ build_bosh_init_manifest() {
 ####################################################
 # multi-call handlers
 
+sha1sum() {
+  if sha1sum 2>/dev/null; then
+    sha1sum "$@"
+  else
+    shasum "$@"
+  fi
+}
+
 cmd_help() {
 	local topic=${1:-usage}
 
--- a/bin/genesis
+++ b/bin/genesis
@@ -166,7 +166,6 @@ must_be_in_an_environment() {
 	fi
 }
 
-
 ####################################################
 # helper functions
 
@@ -1291,14 +1290,6 @@ build_bosh_init_manifest() {
 ####################################################
 # multi-call handlers
 
-sha1sum() {
-  if sha1sum 2>/dev/null; then
-    sha1sum "$@"
-  else
-    shasum "$@"
-  fi
-}
-
 cmd_help() {
 	local topic=${1:-usage}
 
@@ -2005,6 +1996,13 @@ EOF
 	exit 1
 }
 
+sha1sum() {
+	if [[ -z $(command -v sha1sum) ]]; then
+    shasum "$@"
+  else
+    sha1sum "$@"
+  fi
+}
 
 ####################################################
 # multi-call interface
@@ -2125,7 +2123,6 @@ main() {
 		;;
 	esac
 }
-
 
 main $@
 


bin/genesis
```diff
@@ -1291,6 +1291,14 @@ build_bosh_init_manifest() {
...
 # multi-call handlers
 
+sha1sum() {
+  if sha1sum 2>/dev/null; then
+    sha1sum "$@"
+  else
+    shasum "$@"
+  fi
+}
+
```
jhunt: This works great if sha1sum isn't installed; the shell errors because it can't exec the command.

This does not work if sha1sum is installed.

Instead of "succeeding", `sha1sum` blocks forever waiting for input from stdin.  A better way to accomplish this (given that this is bash) is to use the `command` builtin.  Specifically, `command -v sha1sum` works like `which sha1sum` except that (A) it doesn't consider alias functions in the calling environment, and (B) it doesn't result in a fork.  If the binary is not in `$PATH`, `command -v ...` returns no output, but if it *is*, it returns the full path to the binary.  You can check the output with `-n` or `-z` and dispatch accordingly.

johnlonganecker:
I will make the change but it does work fine for me:

```
$ ll
total 112
drwxr-xr-x  3 johnlonganecker  staff   102B Jan 22 11:01 ./
drwxr-xr-x  6 johnlonganecker  staff   204B Jan 21 16:18 ../
-rwxr-xr-x  1 johnlonganecker  staff    55K Jan 22 10:59 genesis
$ sha1sum
^C
$ ./genesis new deployment sha1sum
$ brew uninstall md5sha1sum
Uninstalling /usr/local/Cellar/md5sha1sum/0.9.5... (5 files, 40.2K)
$ sha1sum
bash: /usr/local/bin/sha1sum: No such file or directory
$ shasum
^C
$ ./genesis new deployment shasum
$
```

Compare two new deployments
```
diff sha1sum-deployments/ shasum-deployments/
diff sha1sum-deployments/README.md shasum-deployments/README.md
1c1
< sha1sum Deployments
---
> shasum Deployments
5c5
< sha1sum BOSH deployments, using the format prescribed by the
---
> shasum BOSH deployments, using the format prescribed by the
Common subdirectories: sha1sum-deployments/bin and shasum-deployments/bin
Common subdirectories: sha1sum-deployments/ci and shasum-deployments/ci
Common subdirectories: sha1sum-deployments/global and shasum-deployments/global
```

jhunt:
Test case (for my environment, at least):

```
+11:08:57:504:0:41469m 10.0.1.24 jhunt@pearl.local ~
→  cat test.sh
#!/bin/bash

if sha1sum 2>/dev/null; then
	echo "sha1sum is installed"
else
	echo "not installed"
fi
+11:08:59:505:0:41469m 10.0.1.24 jhunt@pearl.local ~
→  bash test.sh
^C
+11:09:10:506:130:41470m 10.0.1.24 jhunt@pearl.local ~
```

I had to Ctrl-C is to get it to stop.

Generally speaking, _running_ a command to see if it exists on the box is a really bad practice.

johnlonganecker:
makes sense, especially if the command is destructive or should only be run once etc

PR #19, opened 2016-05-20
Support vault prefixes for auto-generating creds

geofffranks:
- Added a default 'meta.vault_prefix' property, so allow
  deployments to reference it when resolving vault keys
- Added support for running scripts as env-setup-hooks,
  so that deployments can have their credentials, or any initial
  setup performed after creating the environment.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -928,10 +928,24 @@ director_uuid: ${director_uuid}
 EOF
 	fi
 
+	export VAULT_PREFIX=secret/${site}/$name/${DEPLOYMENT_NAME}
 	cat <<EOF > ${root}/name.yml
 ----
 name: ${site}-${name}-${DEPLOYMENT_NAME}
+meta:
+  vault_prefix: ${VAULT_PREFIX}
 EOF
+
+	if [[ -d ${DEPLOYMENT_ROOT}/.env_hooks ]]; then
+		pushd ${DEPLOYMENT_ROOT}/.env_hooks
+		for file in *; do
+			if [[ -f ${file} && -x ${file} ]]; then
+				echo "Running env setup hook: ${file}"
+				./${file}
+			fi
+		done
+		popd
+	fi
 
 	refresh_global ${site} ${name}
 	refresh_site   ${site} ${name}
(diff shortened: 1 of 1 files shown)


jhunt:
Looks good.  Waiting to merge until Spruce changes to the `(( vault ... ))` operator are done.

jhunt:
Let me know when this is tested and ready for merge.  I'd like to see version checking on spruce to make sure it is new enough.

geofffranks:
Version checking has been added, and all these have been tested fully

bin/genesis
```diff
@@ -90,6 +133,34 @@ setup() {
 		exit 3
 	fi
 
+	if [[ -f ${DEPLOYMENT_ROOT}/.genesis_deps ]]; then
+		errors=""
+		while IFS='' read -r line || [[ -n "$line" ]]; do
+			cmd=$(echo $line | cut -d ":" -f1)
+			version=$(echo $line | cut -d ":" -f2- | auto_sed 's/^ +//' | auto_sed 's/ +$//')
```
jhunt: My original thought here was to use native string handling routines in bash, to avoid the dep on sed (and the auto_sed function).  Here's what I ended up with needing:

```
cmd=${line%%:*}; cmd="${cmd%"${cmd##*[![:space:]]}"}"; cmd="${cmd#"${cmd%%[![:space:]]*}"}"
ver=${line##*:}; ver="${ver%"${ver##*[![:space:]]}"}"; ver="${ver#"${ver%%[![:space:]]*}"}"
```

(note that I renamed `version` to `ver` to line up the columns)

Not sure how I feel about this.  Maybe a `trim` function?

jhunt: Withdrawn.  However, can we go ahead and use `perl -pi -e` (as below, line 146) instead of sed, to be more uniform?

bin/genesis
```diff
@@ -997,6 +1082,25 @@ create_bosh_init_environment() {
...
+	export VAULT_PREFIX=secret/${site}/$name/${DEPLOYMENT_NAME}
+	cat <<EOF > ${root}/name.yml
+----
+name: ${site}-${name}-${DEPLOYMENT_NAME}
+meta:
+  vault_prefix: ${VAULT_PREFIX}
+EOF
+
+	if [[ -d ${DEPLOYMENT_ROOT}/.env_hooks ]]; then
+		pushd ${DEPLOYMENT_ROOT}/.env_hooks
```
jhunt: Is it important to the environment hooks that `$PWD` be in the hooks directory?  or does it make more sense to be in `$DEPLOYMENT_ROOT`?

Also, do the hook scripts have access to vars like `$DEPLOYMENT_SITE` and `$DEPLOYMENT_ENV`?

geofffranks:
Addressed comments

bin/genesis
```diff
@@ -1082,7 +1086,7 @@ EOF
 		for file in ${DEPLOYMENT_ROOT}/.env_hooks/*; do
 			if [[ -f ${file} && -x ${file} ]]; then
 				echo "Running env setup hook: ${file}"
-				export VAULT_PREFIX DEPLOYMENT_NAME DEPLOYMENT_SITE
+				export VAULT_PREFIX DEPLOYMENT_NAME DEPLOYMENT_SITE DEPLOYMENT_ENVIRONMENT DEPLOYMENT_ROOT ENVIRONMENT_ROOT=${DEPLOYMENT_ROOT}/${DEPLOYMENT_SITE}/${DEPLOYMENT_ENVIRONMENT}
```
jhunt: Do bosh-init environments get the `DEPLOYMENT_ENVIRONMENT` and `DEPLOYMENT_SITE` env vars when you are in the middle of creating a new env?

PR #23, opened 2016-06-07
Allow overriding deployment type when using --template

### LINKED ISSUE #21: `genesis -T` (set type) does not work in conjunction with `--template`
Tried `genesis new deployment -t bosh-init --template bosh my-deployment-name` today, and instead of making it a bosh-init deployment, it made it a regular deployment, though it did pull the templates properly. The `-t` flag should still set the deployment type on the new deployment repo. If any type was provided via upstream templates, `-t` should override that.

geofffranks:
If `-t, --type` is specified while `-T, --template` is
specified, genesis will override the `.deployment` file found
in the upstream templates with the one specified via `-t`.
If not, it will use what was found upstream.

Fixes #21

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -145,7 +145,7 @@ setup() {
 			fi
 		done < ${DEPLOYMENT_ROOT}/.genesis_deps
 		if [[ -n ${errors} ]]; then
-			echo -e ${errors}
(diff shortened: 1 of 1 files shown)


geofffranks:
This will be addressed through #24, #25, #26

PR #37, opened 2016-06-23
Utilize the Genesis Index for Stemcell / Releases

jhunt:
URLs, SHA1 checksums and versions (for "latest") will now be looked up
from the Genesis Index (per the GENESIS_INDEX environment variable).  By
default, this is set to https://genesis.starkandwayne.com/, but can be
overridden (for on-premise instances of the index) or disabled
completely by setting it to the value "no".

If a deployment supplies url file for a stemcell or a release, the trip
to the index will be skipped.  Any releases or stemcells that are
missing will consult the index.

This change also means that "latest" now no longer means "latest
available on your BOSH director", but instead means "latest known to the
index"

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -16,7 +16,7 @@ fi
 WORKDIR=""
 need_a_workdir() {
 	if [[ -z ${WORKDIR} ]]; then
-    WORKDIR=$(mktemp -d -t genesis.XXXXXX)
+		WORKDIR=$(mktemp -d -t genesis.XXXXXX)
 		trap "rm -rf ${WORKDIR}" INT TERM QUIT EXIT
 	fi
 }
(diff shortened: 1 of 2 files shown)


bin/genesis
```diff
@@ -981,6 +987,25 @@ normal_site_metadata() {
...
+	fi
+	if [[ -z ${stemcell_url} ]]; then
+		if [[ -n ${GENESIS_INDEX} ]]; then
+			echo >&2 "Checking ${GENESIS_INDEX} for correct URL/SHA1 of stemcell ${stemcell_name}/${stemcell_version}"
+			stemcell_url=$(    check_index url     stemcell "${stemcell_name}" ${stemcell_version})
+			stemcell_sha1=$(   check_index sha1    stemcell "${stemcell_name}" ${stemcell_version})
+			stemcell_version=$(check_index version stemcell "${stemcell_name}" ${stemcell_version})
+		fi
+	elif [[ -z ${stemcell_sha1} ]]; then
+		stemcell_sha1=$(curl -Lsk ${stemcell_url} | checksum | awk '{print $1}')
```
geofffranks: I don't like this step - seems like it will lead to a bad user experience. I'd rather just error out and tell the person to specify a sha1 themselves. That  way genesis returns quickly, rather than hanging for 5-60 minutes while silently downloading stemcells that could be upwards of a few GB, and potentially exhausting file/memory resources for the download.

ci/release_notes.md
```diff
@@ -0,0 +1,25 @@
...
+
+  This makes it possible for newer BOSH CLIs to transparently
+  upload missing releases and stemcells to your BOSH director,
+  which in turn enables Concourse pipelines to easily roll
+  stemcell and release upgrades through unattended.
+
+  One side-effect of this change is that the "latest" version
+  keyword now no longer means "latest available on the BOSH
+  director", but instead "latest known to the index".
+
```
geofffranks: This concerns me, since 'latest' now means two different things to your manifest, depending on what toolset users are using.  I don't like changing conventions subtly on people, and I'm not sure how well this will go over in an environment under strict change control

jhunt: @dennisjbell had similar concerns, although his complaint mostly centered around when we go to the index -- he advocated talking to the index at `genesis add release` time, where latest would transform to latest-at-that-time and you'd download the URL/SHA1 at that time and not every time you `make deploy`.

I'm open to this.

Thoughts?

bin/genesis
jhunt: Fair.  This may have been a holdover of the old `find_bosh_io_release()` days.  Will remove and error out.

PR #44, opened 2016-06-30
Support running a bosh errand in pipeline

bodymindarts:
Added command `genesis ci smoke-test <errand>` for declaring the errand that should be run as a smoke test after each deployment.
In `ci_repipe_build_environments` `ci_repipe_smoke_test` is called which creates a task config that will execute a bosh errand via `cmd_ci_stage_smoke_test`. The config will get merged together with all other pipeline configuration.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -2381,9 +2381,9 @@ EOF
 
 cmd_ci_smoke_test() {
 	local USAGE="ci smoke-test <errand>"
-	errand=${1}
+	local errand_name=${1}
 
-	if [[ -z ${errand} ]]; then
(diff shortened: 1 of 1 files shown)


jhunt:
Can this be done as an extra line in the main deployment script?

bodymindarts:
It could but I would prefer that git gets updated as soon as the deployment
is completed so that the repo is in sync with what is deployed.
If we only push the changes after an errand there could be a significant
lag between deployment and push making the merge potentially more
complicated.

Also push would get skipped if smoke test fails even though the deployment
completed.

jhunt:
Fair enough,

PR #50, opened 2016-07-01
Genesis image

bodymindarts:
This PR adds a Dockerfile in project root and a job to build said image in the pipeline.
A next step could be to use this image in the pipelines generated by genesis, but even without that this has merit.

Things gained by using a dedicated genesis image:
- easily playing around in an isolated environment with all deps in place
- self contained project including clear documentation of all deps (this could even be improved by using `FROM ubuntu` instead of `FROM concourse/concourse-ci` since all that brings is bosh_cli and it is likely that we want to controll updating the bosh_cli ourselves).
- traceability, easy tracking the history of deps, as requirements / versions change with the tool itself
- reduce responsibility of the starkandwayne/concourse image. The starkandwayne/concourse image has become a 'stuff all things we may need in here' image which makes it impossible to know the relationship between the tools inside it and the projects that use it. When taken to far this will be a source of technical debt when individual projects that use it diverge on their requirements. There is nothing wrong with having a go-to image to get started on a new project pipeline quickly that has only generic requirements. This project however has a focused scope and would benefit from being self contained.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+FROM concourse/concourse-ci
+
+ENV SPRUCE_VERSION=1.5.0 \
+  CF_CLI_VERSION=6.13.0 \
+  VAULT_VERSION=0.6.0 \
+  GENESIS_VERSION=1.5.1
+
+# base packages
+RUN apt-get install -yy curl file
+
+# spruce
+RUN curl -L >/usr/bin/spruce https://github.com/geofffranks/spruce/releases/download/v${SPRUCE_VERSION}/spruce-linux-amd64 \
+      && chmod 0755 /usr/bin/spruce
+
+# jq
+RUN curl -L >/usr/bin/jq http://stedolan.github.io/jq/download/linux64/jq \
+      && chmod 755 /usr/bin/jq
+
+# vault
+RUN curl -L >/tmp/vault.zip https://releases.hashicorp.com/vault/${VAULT_VERSION}/vault_${VAULT_VERSION}_linux_amd64.zip \
+     && unzip /tmp/vault.zip -d /usr/bin/
+
+RUN curl -L >/usr/bin/genesis https://github.com/starkandwayne/genesis/releases/download/v${GENESIS_VERSION}/genesis \
+      && chmod 755 /usr/bin/genesis \
+      && genesis -v
+
--- a/ci/scripts/shipit
+++ b/ci/scripts/shipit
@@ -53,6 +53,7 @@ pushd $REPO_ROOT
 
 mkdir artifacts
 cp bin/${BINARY} artifacts/${BINARY}
+auto_sed Dockerfile "s/GENESIS_VERSION=.*/GENESIS_VERSION=${VERSION}/"
 auto_sed artifacts/${BINARY} "s/^VERSION=.*/VERSION=\"v${VERSION}\"/"
 artifacts/${BINARY} -v | grep ${VERSION}
 popd
--- a/ci/settings.yml
+++ b/ci/settings.yml
@@ -15,3 +15,28 @@ meta:
   slack:
     webhook: FIXME
     channel: "#genesis"
+
+  dockerhub:
+    email: (( vault "secret/pipelines/genesis/dockerhub:email" ))
+    username: (( vault "secret/pipelines/genesis/dockerhub:username" ))
+    password: (( vault "secret/pipelines/genesis/dockerhub:password" ))
+    repository: starkandwayne/genesis
+
+jobs:
+  - name: shipit
+    plan:
+      - (( append ))
+      - put: genesis-image
+        params:
+          build: pushme/git
+          tag: version/number
+          tag_as_latest: true
+
+resources:
+  - name: genesis-image
+    type: docker-image
+    source:
+      email: (( grab meta.dockerhub.email ))
+      username: (( grab meta.dockerhub.username ))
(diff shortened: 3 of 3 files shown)


bodymindarts:
Thinking a step ahead, if/when this image gets used in the generated pipelines we wouldn't have to embed the genesis script inside each genesis managed repository.
We could guarantee that the pipeline will invoke the same version of genesis that the pipeline was created with by specifying the desired image version further improving traceability and reproducibility.

jhunt:
Vault should be in starkandwayne/concourse, so this PR is unnecessary.

bodymindarts:
@jhunt you're comment doesn't address any of the reasons I have named in favor of adding a Dockerfile to this project.

jhunt:
You are fragmenting code for a dubious future optimization

bodymindarts:
Can you be more specific re 'fragmenting' re 'dubious'?

IMO the code is fragmented now (ie spread across multiple repositories.. I am looking to create a coherent project.

jhunt:
Sure,

- easily playing around in an isolated environment with all deps in place

We have not yet had a need to do this regularly.

- self contained project including clear documentation of all deps (this could even be improved by using FROM ubuntu instead of FROM concourse/concourse-ci since all that brings is bosh_cli and it is likely that we want to controll updating the bosh_cli ourselves).

https://hub.docker.com/r/starkandwayne/concourse/

- traceability, easy tracking the history of deps, as requirements / versions change with the tool itself

Not sure what benefit you gain here?  Ought we also track the versions of bash / coreutils / OS / etc?  Our biggest "dep" problem is tracking Concourse changes, which requires regular tweaks to the pipeline YAML configuration.  Image problems have virtually disappeared since we moved to a shared model.

- reduce responsibility of the starkandwayne/concourse image. The starkandwayne/concourse image has become a 'stuff all things we may need in here' image which makes it impossible to know the relationship between the tools inside it and the projects that use it. When taken to far this will be a source of technical debt when individual projects that use it diverge on their requirements. There is nothing wrong with having a go-to image to get started on a new project pipeline quickly that has only generic requirements. This project however has a focused scope and would benefit from being self contained.

The purpose of starkandwayne/concourse *is* to provide a shareable environment.  You can argue the "technical debt" argument if you want, but when we created starkandwayne/concourse, it was in response to the technical debt of having hundreds of little snowflake images that needed updating for new versions of Spruce / CF CLI / etc.

You are adding a special snowflake into the mix by introducing this Docker image, and I don't think the extra snowflakiness is merited.

longnguyen11288:
I'm going to second @jhunt on this one. I don't think having tons of images for each project make sense. We tried doing this for a client to support this flexibility but never had a moment where a pipeline needed a different image because of x, y, z tool. I don't see need of having so many images in which most of them are the same. We should revisit this if `starkandwayne/concourse` image gets bloated from all the changes we have to make all the time.

bodymindarts:
> - easily playing around in an isolated environment with all deps in place
>
> We have not yet had a need to do this regularly.

This is required at least every time someone is learning genesis.

> -self contained project including clear documentation of all deps (this could even be improved by using FROM ubuntu instead of FROM concourse/concourse-ci since all that brings is bosh_cli and it is likely that we want to controll updating the bosh_cli ourselves).
>
> https://hub.docker.com/r/starkandwayne/concourse/

I don't see how that link addresses this point. genesis is not self contained and that Dockerfile is not a clear documentation of genesis deps. Indeed it contains many items that genesis does not depend on and doesn't contain genesis itself.

> -traceability, easy tracking the history of deps, as requirements / versions change with the tool itself
>
> Not sure what benefit you gain here? Ought we also track the versions of bash / coreutils / OS / etc? Our biggest "dep" problem is tracking Concourse changes, which requires regular tweaks to the pipeline. YAML configuration. Image problems have virtually disappeared since we moved to a shared model.

Well if you are questioning the merit of git history and reproducibility that kind of explodes the scope of this discussion.

I think you are taking a too absolute position. The best way to avoid snowflakes in containers is to download the internet because... why not. This is a straw man argument not allowing any room for nuanced discussion. Genesis is a project that wants to offer uniformity in how to deploy environments, creating an image that goes with it increases the uniformity.
There is a difference in having 100s of images (1 per pipeline / deployment environment / app) or having a handful of images explicitly scoped for specific tasks (genesis deployments, building binaries, etc) and having 1 for everything.
Taking an absolute position for no good reason is making discussions around trade offs and application of engineering best practices (explicit over implicit, singe responsibility, segregation + cohesion over coupling) impossible.

What I am reading up to now is:
> we tried one thing and it turned out to be a bad idea so now we are taking an absolutist position regardless of the arguments and experiences of others in the community.

bodymindarts:
> I'm going to second @jhunt on this one. I don't think having tons of images for each project make sense. 

> We tried doing this for a client to support this flexibility but never had a moment where a pipeline needed a different image because of x, y, z tool.

What is _this_ that you tried? Did you create a dedicated tool specifically for a client that contained a dedicated image? Why didn't you use genesis?


>  I don't see need of having so many images in which most of them are the same.

What is _so many_? X > 1? That doesn't leave room for nuanced opinions and discussions of trade off.

> We should revisit this if starkandwayne/concourse image gets bloated from all the changes we have to make all the time.

When would you say an image is bloated? There are a number of tools in https://github.com/starkandwayne/dockerfiles/blob/master/concourse/latest/Dockerfile that genesis doesn't require. Isn't that typically considered bloat.

PR #62, opened 2016-08-17
Disable director lookup when its not available (ie not on the jumpbox)

dennisjbell:
When working locally, or on a jumpbox different from the eventual target jumpbox, the bosh director could be missing.  This patch prevents having to wait the long timeout for 5 failures to reach the bosh director to determine its UUID.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -804,14 +804,23 @@ create_normal_environment() {
 EOF
 	done
 
-	local director_uuid
-	director_uuid=$(bosh status --uuid)
-	if [[ "$?" -eq 0 ]] ; then
-		cat <<EOF > ${root}/director.yml
+	# As a convenience, try to get the BOSH Director UUID, but only if the director in set and reachable.
+	local director_uuid='(( param "Please specify the BOSH Director UUID in '${root}'/director.yml" ))'
+	local director_url
+	local director_address
+	local fetched_director_id
+
+	read director_url director_address < <(bosh target 2>/dev/null | sed -e 's/.* is \(https\{0,1\}:\/\/\(.*\):[0-9]\{1,\}\) (.*)$/\1 \2/')
+	if [[ -n "$director_address" ]] &&\
+		 ping $director_address -c1 -W1 >/dev/null 2>&1 &&\
+		 curl -k -m5 ${director_url}/info >/dev/null 2>&1
+	then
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -799,14 +799,23 @@ create_normal_environment() {
...
-	director_uuid=$(bosh status --uuid)
-	if [[ "$?" -eq 0 ]] ; then
-		cat <<EOF > ${root}/director.yml
+	# As a convenience, try to get the BOSH Director UUID, but only if the director in set and reachable.
+	local director_uuid='(( param "Please specify the BOSH Director UUID in '${root}'/director.yml" ))'
+	local director_url
+	local director_address
+	local fetched_director_id
+
+	read director_url director_address < <(bosh target 2>/dev/null | sed -e 's/.* is \(https\{0,1\}:\/\/\(.*\):[0-9]\{1,\}\) (.*)$/\1 \2/')
```
jhunt: Can we just read ~/.bosh_config via jq?  Do we want to do that?  Is it more or less stable than CLI output?

dennisjbell: since spruce and jq are both requirements for genesis, we can definitely do that.  I was trying to stick to basic bash scripting, but no reason because they have to be there anyway.  Will update.

dennisjbell:
Updated as per the review comments

bin/genesis
```diff
@@ -799,13 +799,24 @@ create_normal_environment() {
...
+		local director_address
+		local fetched_director_id
+
+		director_url=$(spruce json ~/.bosh_config | jq -r ".target")
+		director_address=$(echo $director_url | sed -ne 's/^https\{0,1\}:\/\/\(.*\):[0-9]\{1,\}$/\1/p')
+		if [[ -n "$director_address" ]] &&\
+			 ping $director_address -c1 -W1 >/dev/null 2>&1 &&\
+			 curl -k -m5 ${director_url}/info >/dev/null 2>&1
+		then
+			fetched_director_uuid=$(bosh status --uuid) && director_uuid=${fetched_director_uuid}
```
jhunt: Is this line necessary?  Also, should it match the `local` decl on 806?

dennisjbell: Yeah, that's a remnant of the previous way I was doing things.  New way is cleaner, but left that in.  It works as-is, but I will remove it.

dennisjbell:
Updated to remove remnants of previous changes.

PR #63, opened 2016-08-18
add custom template url support

fearoffish:
Only tested with bitbucket URL's as of now

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1419,7 +1419,11 @@ cmd_new_deployment() {
 		if [[ ${template} != ?*/?* ]]; then
 			template="starkandwayne/${template}"
 		fi
-		template="${template%%-deployment*}-deployment"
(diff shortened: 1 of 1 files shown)


jhunt:
Why not just set a `url` variable and test it for `-z` rather than  mucking about with `custom_url`?

fearoffish:
No reason. That would be a more elegant solution.

fearoffish:
Also, turns out that running `make manifest` later fails, as it checks the index...which knows nothing about the custom template.

jhunt:
Also, also, turns out that @fearoffish likes to confuse issues, and the index problem he was running into was unrelated. ^_^

fearoffish:
Also, also, also, turns out @jhunt is correct. Make sure you don't have 'track' as your version in the release you're using if it isn't in the index.

jhunt:
Waiting on new commits before merging.

PR #68, opened 2016-08-19
New HAVE_TREE check

### LINKED ISSUE #67: Missing check for tree  command
Genesis does not check for availability of tree, but without it,  genesis new (site|env) commands fail during final output.

jhunt:
Since it is not vital to the functionality of genesis to have the `tree`
binary installed locally in $PATH, this commit introduces a check to see
if it is installed, and call sites are updated to only call it if
available.

Fixes #67


jhunt:
@dennisjbell see if this works for your environment.

Issue #69, opened 2016-08-22
Issues with syncing with genesis-index during cmd_build


bodymindarts:
Currently genesis contacts the index during cmd_build(). This has 2 negative effects:
- a lot of commands run cmd_build twice in a row (once redacted once un-redacted) and it is annoying to wait that long twice during the same command flow.
- in a pipeline things can get out of sync. eg when using `track` it is possible for a beta environment to deploy a later version of a release than the alpha environment because the version is resolved while building the manifest.

Suggestion:
Only resolve releases when deploying in alpha environment.
Only resolve stemcells in alpha and beta environment.
All other environments copy the data and don't attempt to resolve to index

jhunt:
I believe the solution to this issue was to not use track with CI pipelines, and to provide a warning when doing so.

PR #70, opened 2016-08-22
Only sync releases and index in alpha/beta environments

bodymindarts:
Addresses issues raised in #69

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1677,7 +1677,7 @@ sync_stemcell_via_index() {
 	for thing in url sha1 version; do
 		v=$(jq -r ".${thing} // \"\"" <${WORKDIR}/index)
 		if [[ -n ${v} ]]; then
-			echo "${v}" > ${stemcells_root}/${thing}
+			echo "${v}" > ${stemcell_root}/${thing}
 		fi
 	done
 }
--- a/bin/genesis
+++ b/bin/genesis
@@ -3277,7 +3277,7 @@ EOF
 		echo "Syncing releases via genesis-index"
 		echo
 		sync_releases_via_index ${WORKING_DIR}/${CI_SITE_ENV}/.global/releases
-		echo "Syncing stemcells via genesis-index"
+		echo "Syncing stemcell via genesis-index"
 		echo
 		sync_stemcell_via_index ${WORKING_DIR}/${CI_SITE_ENV}/.site/stemcell
 		;;
@@ -3289,7 +3289,7 @@ EOF
 		echo
 		cp -a ${WORLD_CHANGES}/${alpha_env}/.global  ${WORKING_DIR}/${CI_SITE_ENV}/.global
 		cp -a ${LOCAL_CHANGES}/${CI_SITE}/site       ${WORKING_DIR}/${CI_SITE_ENV}/.site
-		echo "Syncing stemcells via genesis-index"
+		echo "Syncing stemcell via genesis-index"
 		echo
 		sync_stemcell_via_index ${WORKING_DIR}/${CI_SITE_ENV}/.site/releases
 		;;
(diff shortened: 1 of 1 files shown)


jhunt:
Does this mean that if I'm _not_ using the CI pipelines, I can never use the Genesis Index?

bodymindarts:
The way the current PR is that is correct.
If we want to support looking up genesis-index dynamically during a manual deployment I suggest adding the lookup into the code path during `genesis deploy` rather than `genesis build` or perhaps add a flag `genesis build --sync-index`.
Which ever way you prefer I can add support for in this PR.

jhunt:
@geofffranks might hit me for suggesting it, but `genesis sync (all|site|env)` ?

geofffranks:
I would prefer that a separate sync call not be required to deploy a manifest. Inserting a manual step that must now be run in the middle of an exsiting workflow in order to keep the same behavior is a poor user experience to me.

Can we instead have this sync logic be automatically called when anyone requests a deploy? For `genesis bosh` and other genesis commands, the sync would not be called, and the generated manifest would remain what it was last deployed with (as far as release/stemcell versions are concerned).

bodymindarts:
@jhunt @geofffranks as discussed `genesis deploy` now syncs with index. All other commands reuse cached release/stemcell metadata.
Good to merge?

jhunt:
Can we sync the index on a `genesis build` instead?  Otherwise, the following manual steps are technically incorrect:

```
make manifest
git diff manifests
make deploy
```

bodymindarts:
not syncing during genesis build is the whole point of this PR :-S

jhunt:
I think there are two concerns here, both related to contacting the Genesis index.

1. Gamma environments should use the releases that were vetted in the alpha / beta environments.
2. CLI work with Genesis and the Makefile is annoying, with all of its Genesis syncing.

Those should be addressed to different ways.

(1) should be doable via this patch:

```
diff --git a/bin/genesis b/bin/genesis
index 5ffb16a..b70ad82 100755
--- a/bin/genesis
+++ b/bin/genesis
@@ -2992,10 +2992,11 @@ jobs:
           trigger: true

       - task: ${site_env_name}
         config:
           params:
+            GENESIS_INDEX: no
             CI_SITE_ENV: ${site_env}
         # PUT to git
       - put: git
         params: { repository: out/git, rebase: true }
 EOF
@@ -3054,10 +3055,11 @@ jobs:
             - name: out
           run:
             path: world-changes/bin/genesis
             args: [ci, draft-message]
           params:
+            GENESIS_INDEX:  no
             LOCAL_CHANGES:  local-changes
             WORLD_CHANGES:  world-changes
             CI_SITE_ENV:    ${site_env}
             DRAFT_FILE:     out/notif
       - put: notification
```

That is, set `GENESIS_INDEX` to no in the gamma / omega tasks.

(2) can probably be handled either with a new `genesis sync` command and a modification to the Makefile to do a `genesis sync` before the `genesis build` + `genesis deploy`, or by extracting the sync behavior into a function that gets called in the main dispatch case, once for build (and maybe once for deploy)

bodymindarts:
Your patch doesn't guarantee that beta will use same releases as alpha environment.

geofffranks:
Alternatively, it seems like the issue revolves around using the `track` when used with deployment pipelines. This seems like a potentially bad thing to do with production environments on the whole. 

The reason why stemcells are set to track in the various genesis templates is that I wanted a way to keep them updated while testing things out. Since the genesis ci pipelines can handle stemcell-upgradeing via their own jobs (triggering new stemcells automatically to test in beta environments, and then propagate through), maybe we need to just not use track there. Instead, we could use `latest`, and work-around the failures when `latest` has no release/stemcell uploaded by auto-uploading from the index, if you don't see one present.

bodymindarts:
This has been addressed

PR #72, opened 2016-09-01
Use tagged image in pipeline

bodymindarts:
Move to explicitly tagged genesis image so that pipelines don't break.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM concourse/concourse-ci
+FROM ubuntu
 
 ENV SPRUCE_VERSION=1.5.0 \
   CF_CLI_VERSION=6.13.0 \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 
 ENV SPRUCE_VERSION=1.5.0 \
   CF_CLI_VERSION=6.13.0 \
   VAULT_VERSION=0.6.0 \
   GENESIS_VERSION=1.5.2
 
 # base packages
-RUN apt-get install -yy curl file
+RUN apt-get update \
+      && apt-get install -yy curl file unzip git ruby \
```
geofffranks: @lnguyen  will this get the right version of ruby for bosh_cli, or did we need to do other things to get this working for GE?

geofffranks:
What about just making the pipeline configs use meta.docker_image, and default that to starkandwayne/concourse, overriding as necessary down the road? We have a bunch of non-genesis pipelines with all the same requirements as genesis, so diverging now duplicates the maintenance of that featureset.

If down the road we need to diverge a genesis image, we can base it off starkandwayne/concourse, and add the genesis-specific items?

Or if you prefer, we have a starkandwayne/concourse base image, and starkandwayne/concourse-bosh built off of it, adding all the genesis/boshrelease concourse dependencies?

jhunt:
@bodymindarts any thoughts on @geofffranks' comments?

bodymindarts:
Sounds good to me. The only concern I have is regarding the versioning. Can
we keep starkandwayne/concourse versions in sync with Genesis versions? Or
is it possible to guarantee that Genesis would not pick up a broken image
or one with incompatible deps?

jhunt:
Closing this old PR.

PR #78, opened 2016-09-26
Update release/stemcell upload/version checking behavior

### LINKED ISSUE #77: Define how versions should be managed + Genesis Index should be contacted
Right now, there isn't any formal documentation about when Genesis Index is contacted, or how things like `track`, `latest`, and `1.3.4` versions should behave. This makes it hard for people to anticipate what genesis will do under certain circumstances. We should add some, to set expectations.

geofffranks:
Brings Genesis behavior inline with the rules set forth in #77.

Genesis will now auto-upload releases (if they don't exist), and no
sha1/url are present, when versions are latest, or a specific number.

If track is specified, it will behave the same.

Fixes #77.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,3 +82,90 @@ If you have privileged access and wish to install Genesis for all
 users, you can put it in `/usr/local/bin` or `/usr/bin`
 
 [spruce]: https://github.com/geofffranks/spruce
+
+# Managing BOSH Release/Stemcell Versions with Genesis
+
+Genesis offers the following commands for manipulating release
+and stemcells globally, and across sites:
+
+| Command | Purpose |
+| ------- | ------- |
+| `genesis add release <release-name> [<version>]` | Adds the release to global/releases, at the specified version, or 'latest' if not specified |
+| `genesis use release <release-name>` | Adds the release to site/releases, so it gets included in environments of this site ||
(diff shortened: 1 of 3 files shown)


geofffranks:
Cases to test:
1. Verify when not deploying, that director isn't contacted and releases aren't uploaded
2. Verifying when deploying that director is checked for release being there (when using latest)
2a. upload happens when release is not present
2b. upload doesn't happen when release is present
3. Verify when deploying taht the director is checked for release being there (when using version)
3a. upload happens when release is not present
3b. upload doesn't happen when release is present
4. Verify that when using track, index is always checked.
5. Verify error on can't talk to bosh
6. Verify error on not in index
7. Verify error on index disabled, but needed to deploy


Configuration:
stemcell - latest
release toolbelt - track
release jumpbox - specific version
release jehosephat - latest


Test Plan:
1. make manifest
- should only hit track for toolbelt
2. make deploy
- will fail on jehosephat for not being in index
3. turn off bosh-lite, deploy
- get error
4. switch-directors, deploy
- get error
5. disable index
- get error
5. nominal director, make deploy
- should upload stemcell, jumpbox, toolbox
6. secondary make deploy (after everything uploaded)
- should not upload anything, check index for track
7. disable index, make refresh deploy
- should fail on track

Test Results:
1. passed
2. passed
3. passed
4. passed
5. passed
6. passed
7. passed

PR #95, opened 2016-11-04
Add quotes to versions in manifest

### LINKED ISSUE #94: Stemcell version xxx.10 isn't being put into final manifest properly.
the problem is when merging with spruce:
```
$ cat test.yml
meta:
  version: 123.10

stemcell:
  version: (( grab meta.version ))

$ spruce merge test.yml
meta:
  version: 123.1

stemcell:
  version: 123.1
```
This breaks uploading stemcells. It looks like spruce treats .10 as a float and leaves off the 0 when outputting the result. Perhaps we can move to a string based handling of versions.

bodymindarts:
Fixes #94

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1001,15 +1001,6 @@ preflight_normal_deployment() {
 
 	ensure_present stemcell ${stemcell_name} ${stemcell_version}
 
-	local stemcell_url=""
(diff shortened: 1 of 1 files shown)


bodymindarts:
Added another commit to fix regression in smoke_test code

PR #101, opened 2016-11-24
simple custom template url support

fearoffish:
Add support for using a custom URL for templates not on github

Finally got back to finishing this, I think this is the method that @jhunt was hoping for?

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1678,6 +1678,9 @@ cmd_new_deployment() {
 	done
 
 	if [[ -n ${template} ]]; then
+		if [[ ${template} =~ (http(s)?):\/\/ ]]; then
+			custom_template=true
+		fi
 		if [[ ${template} != ?*/?* ]]; then
 			template="starkandwayne/${template}"
 		fi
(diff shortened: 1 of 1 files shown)


geofffranks:
Tested this out, and it seems to break things like `genesis new deployment --template cf` (should use `https://github.com/starkandwayne/cf-deployment`) or `genesis new deployment --template geofffranks/test` (should use `https://github.com/geofffranks/test-deployment`)

bin/genesis
```diff
@@ -1701,7 +1704,11 @@ cmd_new_deployment() {
 	fi
 
 	if [[ -n ${template} ]]; then
-		local url=https://github.com/${template}
+		if [[ ! -z custom_template ]]; then
```
geofffranks: Did you mean `! -z ${custom_template]`?

jhunt:
*bump*

fearoffish:
I've made the changes and pushed to fearoffish/custom-template-url

jhunt:
@geofffranks can you re-review?

geofffranks:
Closing, since genesis-v2 obviates this

PR #103, opened 2016-11-30
Fixes

bodymindarts:
2 minor fixes:
1) Exit earlier in pipeline script when BOSH director is unavailable.
2) Remove url and sha1 files when updating a release version to force contact with genesis-index on next deploy.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -2025,6 +2025,7 @@ cmd_set_release() {
 
 	setup
 	mkdir -p ${DEPLOYMENT_ROOT}/global/releases/${release}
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -2025,6 +2025,7 @@ cmd_set_release() {
 
 	setup
 	mkdir -p ${DEPLOYMENT_ROOT}/global/releases/${release}
+	rm -rf ${DEPLOYMENT_ROOT}/global/releases/${releases}/*
```
geofffranks: This should probably be `${release}` not `${releases}`?

bodymindarts: Yup fixed here: 5d44696163e571006efce1a83e172818f9ed9fac

PR #107, opened 2016-12-06
Ci levels

### LINKED ISSUE #105: Feature support for more stages in pipeline
We would like to support alpha / beta / stage-3 / stage-4 pipelines.

bodymindarts:
This PR enables the possibility of configuring pipelines with more than 3 stages.

The command genesis ci parent <site/env> <parent_env> will configure genesis to consider <parent_env> the predecessor environment on the next genesis ci repipe command.

This only effects environments that were previously of the gamma / omega flavours and is completely backwards compatible.

Closes #105.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -2741,6 +2741,30 @@ EOF
 	done
 }
 
+cmd_ci_parent() {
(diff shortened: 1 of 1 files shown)


bodymindarts:
This code is running and tested with client.

PR #109, opened 2017-01-04
genesis now only check version for itself

xiujiao:
Now genesis will only check version :)

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -139,7 +139,13 @@ setup() {
 		while IFS='' read -r line || [[ -n "$line" ]]; do
 			cmd=$(echo $line | perl -pe 's|^(.*?):.*$|$1|')
 			ver=$(echo $line | perl -pe 's|^.*?: *(.*)$|$1|')
-			if [[ -z $(command -v ${cmd}) ]]; then
+			if [[ ${cmd} == "genesis" ]] ; then
+				if [[ ${ver} != ${VERSION} ]] ; then
+					errors="This genesis deployment depends on ${cmd} version ${ver}, "\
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -139,7 +139,13 @@ setup() {
 		while IFS='' read -r line || [[ -n "$line" ]]; do
 			cmd=$(echo $line | perl -pe 's|^(.*?):.*$|$1|')
 			ver=$(echo $line | perl -pe 's|^.*?: *(.*)$|$1|')
-			if [[ -z $(command -v ${cmd}) ]]; then
+			if [[ ${cmd} == "genesis" ]] ; then
+				if [[ ${ver} != ${VERSION} ]] ; then
```
dennisjbell: Pinning to a specific version may be too restrictive; do we want to support a minimum version ie (>= v1.2.3)

xiujiao: I thought about the same thing when I worked on it. The reason I did not do support a minimum version is because: for example, we specify `genesis 1.5.5` in `.genesis_deps` in a deployment repo, while actually we have installed genesis 1.6.0. then `genesis` will tell us, hi we require `1.5.5` while you have `1.6.0`. In this case, it reminds us to update `.genesis_deps` to use new version instead of sticking to the old version. I saw that we did not always keep `.genesis_deps` configured with latest version for all the repos. This strict version check should help with it.

jhunt: We already support minimum version checking.  The change here should be to just use the value of `$VERSION` instead of running `genesis -v`.  We should still do the comparison down below.

xiujiao:
@jhunt Hi, I made the changes as you requested

PR #115, opened 2017-01-19
Add CI support for waiting on active BOSH tasks

geofffranks:
When CI deploys, you can configure the pipeline to
wait for existing BOSH tasks to finish before deploying,
by setting `meta.pause_for_existing_bosh_tasks` in `ci/settings.yml`

The CI task will wait up to 100 seconds for BOSH to finish tasks
before bailing out.


bin/genesis
```diff
@@ -4445,6 +4452,49 @@ EOF
...
+
+		if [[ ${CI_WAIT_FOR_CLEAN_BOSH} == "true" ]]; then
+			i=1
+			attempts=10
+			wait=10
+			echo "Checking if BOSH has any tasks running... (for up to $(( $attempts * $wait )) seconds)"
+
+			while (( $i <= $attempts ))
+			do
+				tasks=$(ask_bosh '/tasks?state=queued,processing,cancelling&verbose=2')
```
geofffranks: This asks the wrong bosh for tasks. It asks the bosh which will deploy the current deployment, rather than the bosh of the current deployment.

bin/genesis
```diff
@@ -4443,6 +4455,52 @@ EOF
...
+				fi
+
+				running=$(echo "${tasks}" | jq -r '. | length')
+				if [[ $? != 0 ]]; then
+					echo "FATAL: Unexpected task output from BOSH. Cowardly refusing to deploy"
+					echo "Unexpected response:"
+					echo
+					cat << EOF
+$tasks
+EOF
```
dennisjbell: why not just `echo "$tasks"` ?

geofffranks: I didn't want to worry about any special characters (like newlines) that may have been inside $tasks not being printed/escaped properly through just an `echo` call.

geofffranks: Seems `echo "tasks"` would work just as well

geofffranks:
The purpose of unsetting the DIRECTOR_URL is to work around any caching that may have been done in `ask_bosh`/`need_a_bosh` calls prior/after querying the to-be-deployed bosh for its task status. if DIRECTOR_URL is set, `need_a_bosh` won't re-generate the URL based off the currently targeted bosh

PR #116, opened 2017-01-23
now draft message tells which deployment is ready to go

xiujiao:


bodymindarts:
Its working in our env

xiujiao:
@bodymindarts Thank you :)

bin/genesis
```diff
@@ -4570,9 +4570,8 @@ EOF
 
 cmd_ci_draft_message() {
 	[[ -z $1 ]] || bad_usage "ci draft-message"
-
 	(
-		echo "[${CI_SITE_ENV}] Deployment changes are ready to go!"
+		awk '/^name: /{print "[" $2 "] Deployment changes are ready to go!"}' "${LOCAL_CHANGES}/${CI_SITE_ENV}/name.yml"
```
jhunt: What happens if `$.name` is a Spruce operator?

xiujiao: @jhunt Do you mean in `name.yml`, the name is specified by Spruce operator? 
In the current `genesis` version, name is got  by`name: ${site}-${name}-${DEPLOYMENT_NAME}, so in the `name.yml`, I can retrieve the name correctly without worrying about the case it is specified by Spruce operator. If in future, it does use Spruce operator to specify name, I can get name from `manifest.yml`, but the downside is that we have to generate `manifest.yml` for prod and push it before we configure the pipeline.
Any other good idea about this?

geofffranks:
Closing, since genesis-v2 obviates this

PR #133, opened 2017-03-29
added support for multiple ci environments in repipe

lnguyen:


bin/genesis
```diff
@@ -3096,15 +3096,6 @@ ci_update() {
 	fi
 }
 
-ci_pipeline_yaml() {
```
dennisjbell: why'd you get rid of this?

lnguyen: cause we have two functions with same name

jhunt: kudos on picking the right one to remove.  the remaining definition is actually correct 🤣

jhunt:
LGTM :+1:

PR #166, opened 2017-08-24
Extract all hook-related files when running subkit/param hooks

geofffranks:


geofffranks:
This should fix the issue seen in https://github.com/genesis-community/cf-genesis-kit/issues/2

PR #170, opened 2017-09-28
Add "genesis ci" command

xiujiao:
* Add `genesis ci` command, so user can simply answer questions to automatically generate pipeline template 

* Some bug fix in `prompt_for_line`

* Add sub `is_valid_uri` with very basic validation

* Tests code are not included in this PR

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -152,7 +152,9 @@ sub trace(@) {
 }
 
 sub error {
-	printf STDERR csprintf(@_) . "\n";
+	my @err = @_;
+	unshift @err, "%s" if $#err == 0;
+	print STDERR csprintf(@err) . "\n";
 }
 
 sub online {
@@ -761,22 +763,41 @@ sub __prompt_for_line {
 		if ($validation eq "url") {
 			$validate = sub() {
 				return "" if (is_valid_uri($_[0]));
-				return ($_[1] ? $_[1] :"$_[0] is not a valid URL\n");
+				return ($_[1] ? $_[1] :"$_[0] is not a valid URL");
 			}
 		} elsif ($validation eq "port") {
 			$validate = sub() {
 				return "" if ( ($_[0] =~ m/^\d+$/) && ($_[0] >= 0 ) && ($_[0] <= 65535) );
-				return ($_[1] ? $_[1] :"$_[0] is not a valid port\n");
+				return ($_[1] ? $_[1] :"$_[0] is not a valid port");
 			}
-		} elsif ($validation =~ m/^\/(.*)\/(i?m?s?)$/) {
-			my $__vre;
-			# safe because the only thing being eval'ed is the optional i,s, or m
-			eval "\$__vre = qr/\$1/$2"
-				or die "Error compiling param regex: $!";
+		} elsif ($validation =~ m/^(-?\d+(?:.\d+)?)-(-?\d+(?:.\d+)?)$/) {
+			my ($__min,$__max) = ($1,$2);
 			$validate = sub() {
-				return ($_[0] =~ $__vre ? "" : ( $_[1] ? $_[1] : "Does not match required pattern"));
+				return "" if (($_[0] =~ m/^\d+$/) && ($_[0] >= $__min ) && ($_[0] <= $__max));
+				return ($_[1] ? $_[1] : "$_[0] expected to be between $__min and $__max");
 			}
-		} elsif ($validation =~ m/^((^|,)[^,]+){2,}$/) {
+		} elsif ($validation =~ m/^(!)?\/(.*)\/(i?m?s?)$/) {
+			my $__vre;
+			my $__negate = ($1 && $1 eq "!");
+			# safe because the only thing being eval'ed is the optional i,s, or m
+			eval "\$__vre = qr/\$2/$3"
+				or die "Error compiling param regex: $!";
+			$validate = $__negate ? sub() {
+				return ($_[0] !~ $__vre ? "" : ( $_[1] ? $_[1] : "Matches exclusionary pattern"));
+			} : sub() {
+				return ($_[0] =~ $__vre ? "" : ( $_[1] ? $_[1] : "Does not match required pattern"));
+			};
+		} elsif ($validation =~ m/^(!)?\[([^,]+(,[^,]+)*)\]$/) {
+			my @__vlist = split(",", $2);
+			my $__negate = ($1 && $1 eq "!");
+			$validate =  sub() {
+				my $needle=shift;
+				my @matches = grep {$_ eq $needle} @__vlist;
+				return $__negate ?
+					(scalar(@matches) == 0 ? "" : ($_[1] ? $_[1] : "Cannot be one of ".join(", ",@__vlist))):
+					(scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)));
+			}
+		} elsif ($validation =~ m/^((^|,)[^,]+){2,}$/) { # Depricated list match 2 or more
 			my @__vlist = split(",", $validation);
 			$validate = sub() {
 				my $needle=shift;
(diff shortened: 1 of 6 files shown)


bin/genesis
```diff
@@ -758,7 +758,23 @@ sub __prompt_for_line {
...
+			$validate = sub() {
+				if (is_valid_uri($_[0])) {
+					return "";
+				} else {
+					return ($_[1] ? $_[1] :"$_[0] is not a valid URL\n");
+				}
+			}
+		} elsif ($validation eq "port") {
+			$validate = sub() {
+				if ( ($_[0] =~ /^\s*[\+\-]?\d+\s*$/) && ($_[0] >= 0 ) && ($_[0] <= 65535) ) {
```
jhunt: You don't need to escape '+' or '-' inside of the character class brackets - they lose their special meaning as soon as you cross into `[...]` territory.

That being said, if this is to validate ports, why are you even allowing a sign on the number?  port "-80" isn't a real port.  I would drop the sign and just check against `\d+`

bin/genesis
```diff
@@ -758,7 +758,23 @@ sub __prompt_for_line {
 	# being an optional error message
 	my $validate;
 	if (defined($validation)) {
-		if ($validation =~ m/^\/(.*)\/(i?m?s?)$/) {
+		if ($validation eq "url") {
+			$validate = sub() {
+				if (is_valid_uri($_[0])) {
+					return "";
```
jhunt: It's more perl-y (and I believe @dennisjbell  will agree, to phrase this as:

```
return "" if is_valid_uri($_[0])
```

bin/genesis
```diff
@@ -1497,7 +1533,7 @@ sub parse_pipeline {
 	my $src = $pipeline->{pipeline}{layouts}{$layout}
 		or die "No such layout `${layout}'\n";
 
-	# our internal representatio
+	# our internal representation
```
jhunt: Is this relevant to the addition of a `ci` sub-command?

bin/genesis
```diff
@@ -771,12 +787,12 @@ sub __prompt_for_line {
...
 				my @matches = grep {$_ eq $needle} @__vlist;
-				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)))
+				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)));
 			}
 		} elsif ($validation eq "vault_path") {
-			my $validate = sub() {
+			$validate = sub() {
 				system "safe exists $_[0]";
-				return ( $? eq 0 ? "" : ($_[1] ? $_[1] :"$_[0] not found in vault\n"))
+				return ( $? eq 0) ? "" : ($_[1] ? $_[1] :"$_[0] not found in vault\n");
```
jhunt: These appear to be unrelated bug fixes, and should really go on their own PR.

bin/genesis
```diff
@@ -771,12 +787,12 @@ sub __prompt_for_line {
 			$validate = sub() {
 				my $needle=shift;
 				my @matches = grep {$_ eq $needle} @__vlist;
-				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)))
+				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)));
 			}
 		} elsif ($validation eq "vault_path") {
-			my $validate = sub() {
+			$validate = sub() {
```
jhunt: These appear to be unrelated bug fixes, and should really go on their own PR.

bin/genesis
```diff
@@ -771,12 +787,12 @@ sub __prompt_for_line {
 			$validate = sub() {
 				my $needle=shift;
 				my @matches = grep {$_ eq $needle} @__vlist;
-				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)))
+				return (scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)));
```
jhunt: These appear to be unrelated bug fixes, and should really go on their own PR.

bin/genesis
```diff
@@ -3655,6 +3691,296 @@ sub {
...
+
+	$pipeline{git} = prompt_for_git();
+
+	$pipeline{notifications} = prompt_for_notifications();
+
+	$pipeline{locker} = prompt_for_locker();
+
+	$pipeline{vault} = prompt_for_vault();
+
+	$pipeline{stemcells} = prompt_for_block("What is(are) the stemcell(s) you need for all your deployments in this pipeline? \nOne stemcell per line, for example you can input the bosh-lite stemcell like the following example and then hit enter.\nbosh-lite: bosh-warden-boshlite-ubuntu-trusty-go_agent\nNote:\"bosh-lite\" is the short name you want to call this stemcell which will be used to configure stemcell for your \"boshes\" later; \"bosh-warden-boshlite-ubuntu-trusty-go_agent\" is the stemcell name from 'bosh.io'.\n");
```
jhunt: Is there a better way to format this code, perhaps via a heredoc?  Long lines are very hard to read.

xiujiao:
@jhunt
 I will add test next.

yeah, feedback about prompting messages are very appreciated, I did have a hard time to say everything accurately in a simple way.

The motivations are that 1) all other steps are automated, while only this step requires writing by hand;  2) for people who are new to genesis, they have to go to read `pipeline.md` to figure out what should be configured and genesis code is not always consistent with doc `pipeline.md`; 3) from user perspective, it will be nice they only need run several genesis command and answer some questions, then the whole deployment and pipelines are set up.
I am not sure how much simplifying YAML structure will help with the above, but definitely open to better ideas always.

bin/genesis
xiujiao: I think you are right. This is first time I code in perl, so it is easy to be perl-y, or perl-c :)
I will fix this

bin/genesis
xiujiao: Thanks, will modify

bin/genesis
xiujiao: this function `__prompt_for_line` is called by `genesis ci` command, so I just put those bug fix in the same PR. I did put those bug fix in its own commit to separate from the main code for adding `ci` feature.

bin/genesis
xiujiao: this function `__prompt_for_line` is called by `genesis ci` command, so I just put those bug fix in the same PR. I did put those bug fix in its own commit to separate from the main code for adding `ci` feature.

bin/genesis
xiujiao: Not really directly, it is one tiny typo I noticed, so I just fixed it without a separate PR

bin/genesis
xiujiao: I will do some research on it , thanks for your feedback !!

bin/genesis
```diff
@@ -761,22 +763,41 @@ sub __prompt_for_line {
...
+			my $__negate = ($1 && $1 eq "!");
+			$validate =  sub() {
+				my $needle=shift;
+				my @matches = grep {$_ eq $needle} @__vlist;
+				return $__negate ?
+					(scalar(@matches) == 0 ? "" : ($_[1] ? $_[1] : "Cannot be one of ".join(", ",@__vlist))):
+					(scalar(@matches) != 0 ? "" : ($_[1] ? $_[1] : "Expecting one of ".join(", ",@__vlist)));
 			}
-		} elsif ($validation =~ m/^((^|,)[^,]+){2,}$/) {
+		} elsif ($validation =~ m/^((^|,)[^,]+){2,}$/) { # Depricated list match 2 or more
```
dennisjbell: Depricate -> Deprecate

bin/genesis
```diff
@@ -66,6 +66,21 @@ sub envdefault {
 	return defined $ENV{$var} ? $ENV{$var} : $default;
 }
 
+sub valid_envs {
+	my (%envs, $json);
+	foreach (glob "*.yml") {
+		eval {$json = decode_json(qx{spruce json "$_"});};
```
dennisjbell: Redirect STDERR to STDOUT for spruce

PR #196, opened 2017-12-22
Enforce hierarchial naming of environments to promote pipelines, create top-level env file.

### LINKED ISSUE #188: on `genesis init` prompt for base yml file to stuff common stuff
For pipelines to be most effective, genesis env files should share as much as possible with the parent files (ie company.yml is the parent of company-env.yml).  Right now, we rely on manually extracting kit information to the parent file after creating a new env using `genesis new`.

One of the most common things to put in the top parent file is the kit information.  This should be done on `genesis init` to facilitate and promote this proper work flow in support of pipelines and their propagation of cached parent files.  We would just need to prompt for the top parent and write the kit information to it.

Further enhancements could prevent or warn on `genesis new` if the env name does not build off an existing parent.

dennisjbell:
Genesis provides pipelines to propagate changes through CI phases, but by default, there is no innately shared values between the different environments.  Or is there...?

The environment files are actually based on a hierarchical structure using dashes to separate the files.  This means that corp-zone-env.yml env file actually inherits from corp.yml and corp-zone.yml, if they exist.  Furthermore, these ancestral files propagate through the pipeline, so they need to pass previous tasks before they get to later tasks.  What does this mean?  It means that if you increment the kit version in corp.yml, then it has to pass corp-dev before it gets applied to corp-prod (assuming that's how you configured your pipeline)

To this end, this PR does two things:  It automatically creates the top-level file, containing the kit name, version and subkits, and second, it doesn't let you create flat envs (env names that don't contain a dash) to promote awareness and usage of these hierarchies.

Fixes #188, though not in the way it was described.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -168,6 +168,15 @@ sub trace(@) {
 	print STDERR "\n";
 }
 
+sub same($$) {
+	my ($a,$b) = @_;
+	die "Arguments are not arrays" unless ref($a) eq 'ARRAY' && ref($b) eq 'ARRAY';
+	return 0 unless scalar(@$a) == scalar(@$b);
+	return 0 unless join(',',map {length} @$a) eq join(',',map {length} @$b);
+	return 0 unless join("\0", @$a) eq join("\0",@$b);
+	return 1;
+}
+
 sub error {
 	my @err = @_;
 	unshift @err, "%s" if $#err == 0;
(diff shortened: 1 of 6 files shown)


bin/genesis
```diff
@@ -4133,7 +4185,7 @@ sub prompt_for_git {
 	my $gitremote = qx(git config --get remote.origin.url);
 	if ($gitremote) {
 		chomp($gitremote);
-		($user,$host,$org,$repo) = $gitremote =~ qr/^(?:ssh:\/\/([^@]+)@)?([^\/:]*)[\/:](?:(.*)\/)?([^\/]+)$/
+		($user,$host,$org,$repo) = $gitremote =~ qr/^(?:(?:ssh:\/\/)?([^@]+)@)?([^\/:]*)[\/:](?:(.*)\/)?([^\/]+)$/
```
jhunt: Is this related?

bin/genesis
```diff
@@ -4548,6 +4600,11 @@ sub boolean_to_yaml {
 sub generate_ci_template_file() {
 	my ($pipeline, $name) = @_;
 
+	use Data::Dumper;
+	open(my $fh, ">", "/tmp/ci-data.out");
+	print $fh Dumper($pipeline);
+	close $fh;
+
```
jhunt: We should probably remove this debugging.

bin/genesis
```diff
@@ -405,22 +414,60 @@ sub new_environment {
...
-	print $fh <<EOF;
+	(my $parent = $file) =~ s/-.*\.yml/.yml/;
+
+	my %existing_info;
+	if (-e $parent) {
+		%existing_info = %{LoadFile($parent)};
+		explain "Using #C{$parent} file as base config.";
+		#open my $fh, ">",  $parent.'-debug';
+		#use Data::Dumper;
+		#print $fh Dumper(\%existing_info);
```
jhunt: We should probably remove this (commented-out) debugging.

PR #197, opened 2017-12-22
Vault prefix usage and generation refactor

dennisjbell:
This fixes the following issues with vault_prefix usage and generation:
- Extracts/standardizes vault slug generation from env name, and applies it consistently
- Uses params.vault if present in the env file in `genesis secrets`
- Corrects erroneous conversion of dashes to slashes of the deployment name to create the vault prefix in `genesis secrets`
- Allows users to specify an alternate vault_prefix on `genesis new`, so that secrets generated by it go to the needed location.

Minor issues also resolved:
- Drops support for detecting environment files with .yaml as extension (they wouldn't be picked up for merge inclusion anyways)
- Improves error handling of `genesis secret` when user forgets to specify environment file
- More consistent naming of vault prefix variable in `genesis ci` supporting subroutine

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -3271,7 +3271,8 @@ sub process_kit_params {
 	my %opts = @_;
 	my @answers;
 	my $resolveable_params = {
-		"params.vault_prefix" => $opts{vault_prefix},
+		"params.vault_prefix" => $opts{vault_prefix}, # for backwards compatibility
+		"params.vault" => $opts{vault_prefix},
 		"params.env" => $opts{env}
 	};
 	for my $subkit ("base", @{$opts{subkits}}) {
--- a/bin/genesis
+++ b/bin/genesis
@@ -528,7 +528,7 @@ EOF
 
 sub kit_yaml_files_in {
 	my ($env, $dir, @subkits) = @_;
-	$env =~ s/\.ya?ml$//;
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -528,7 +534,7 @@ EOF
 
 sub kit_yaml_files_in {
 	my ($env, $dir, @subkits) = @_;
-	$env =~ s/\.ya?ml$//;
+	$env =~ s/\.yml$//;
```
jhunt: Why no love for `.yaml`?

bin/genesis
```diff
@@ -4001,13 +4013,14 @@ sub {
...
 		vault=s
+		prefix=s
 		secrets!
 		kit|k=s
 	/);
 	usage(1) if @_ != 1;
 
 	my ($name) = @_;
-	$name =~ s/\.ya?ml$//;
+	$name =~ s/\.yml$//;
```
jhunt: Did `.yaml` offend your honor? 😀

bin/genesis
dennisjbell: https://github.com/starkandwayne/genesis/blob/master/bin/genesis#L629-L656

jhunt: Annoying.

PR #205, opened 2018-01-19
Deprecate subkits

dennisjbell:
In anticipation of supporting upstream repos that use ops files (namely cf-deployment and bosh-deployment), we have expanded on the concepts around subkits and renamed them features.

This PR handles the renaming of the concept from subkit to feature in yaml files and directory structure, while still supporting but deprecating the usage of subkit in existing kits.  A follow-up PR will add the ability to go beyond using `feature` (formally known as `subkit`) directories to encapsulate the files that provide that feature.


t/kit-validation.t
```diff
@@ -18,9 +18,11 @@ run_fails "genesis new failures-galore > errors 2>&1", 255, "Invalid kits fail t
 eq_or_diff get_file("errors"), <<EOF, "Invalid kits generate expected error messages";
 Generating new environment failures-galore...
 
-Using dev/ (development version) kit...
+Using local development kit (./dev)...
 
-The following errors have been encountered validating the dev/latest kit:
+\e[1;33mDEPRICATED:\e[0m subkits are depricated in favour of features.  Please update your kit.
```
jhunt: DEPRECATED has 3 E's

dennisjbell: Can I claim that it's the Canadian spelling?

PR #208, opened 2018-01-23
Localize BOSH cred usage to prevent new X509 error

dennisjbell:
A previous fix solved the problem when trying to get stemcells if the
BOSH config didn't match the URL used by `genesis ci`, but in doing so,
polluted the ENV space for further commands.  This make those changes
local to just where its needed when retrieving the stemcells.


xiujiao:
LGTM

PR #211, opened 2018-01-25
Don't honor BOSH_ENVIRONMENT

jhunt:


dennisjbell:
Adding these notes for those that follow and want to know why this was done.

While we continue to support GENESIS_BOSH_ENVIRONMENT, we have elected to not support the normal BOSH_ENVIRONMENT in order to safeguard your deployments.  The correct way to specify the bosh directory for an environment is by specifying the actual URL in your environment yml, followed by explicitly stating the bosh alias, and finally relying on the convention that your environment is named the same as your bosh alias for the desired director.  We have found that the likelihood of your BOSH_ENVIRONMENT being set to something else due to past actions is significantly higher than it being set to what you actually want to use, and this was causing new env files and deployment to erroneously target the previously set BOSH_ENVIRONMENT, and it does this without informing the user.

We could have kept the BOSH_ENVIRONMENT for deployment (or anything other than new) and just lowered its priority, but since the environment name for the yaml file is ALWAYS specified, and it would have to take precedence over BOSH_ENVIRONMENT, this would have no effect.

In the end, we felt it clearer and safer to specify the environment using the --environment option in the cases where you really wanted to override the defaults.

Issue #218, opened 2018-01-31
Genesis does not recognize Vault's changed version string


norman-abramovitz:
Vault version 0.9.3

genesis init --kit bosh-init
Unrecognized version string 'Vault v0.9.3 ('5acd6a21d5a69ab49d0f7c0bf540123a9b2c696d')' for vault

dennisjbell:
Odd that it does recognize it for 0.9.1:
Vault v0.9.1 ('87b6919dea55da61d7cd444b2442cabb8ede8ab1')

I'll look into this.

norman-abramovitz:
Could be me since I am flying blind on figuring out how to use Genesis.

jhunt:
What version of Genesis are you using, @norman-abramovitz ?

norman-abramovitz:
Genesis v2.4.1

PR #219, opened 2018-01-31
Ignore garbage prepended to vault version string

dennisjbell:
Vault added ANSI control sequence to the start of their version string from 0.9.2 onward, which confuses the version detection code in genesis.  This resolves the problem by ignoring it :)


dennisjbell:
Fixes #218

PR #222, opened 2018-02-08
Tag the resources as necessary

jhunt:
In tagged environments, not all BOSH directors are necessarily
accessible from the untagged worker set, so we need to tag the resources
so that they check from the appropriate workers.


jhunt:
@dennisjbell I believe I am finished tweaking this PR.

PR #226, opened 2018-02-20
Fixup prompt for ip

dennisjbell:
Adds ip validation to prompt-for line.

Also fixes issues with hooks/new script that use `set -eu` and adds the missing --echo option for the secret-line prompt.

### DIFF
--- a/t/param-asking.t
+++ b/t/param-asking.t
@@ -349,6 +349,53 @@ expect_ok "URL parsing - finish up", $cmd, [
   }
 ];
 
+
+expect_ok "IP parsing - initial prompt", $cmd, [
+  "IP Address[\r\n]{1,2}>", sub {
+    $_[0]->send("mysite.example.com\n");
+  }
+];
+expect_ok "IP parsing - used domain name", $cmd, [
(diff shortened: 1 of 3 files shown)


t/param-asking.t
```diff
@@ -349,6 +349,53 @@ expect_ok "URL parsing - finish up", $cmd, [
...
+];
+expect_ok "IP parsing - too many octets", $cmd, [
+  "Specify the path", sub {
+    fail "Invalid IP address not correctly identified: too many octets";
+  }], [
+  "123.4.5.6.78 is not a valid IPv4 address[\r\n]{1,2}>", sub {
+    $_[0]->send("123.4.5.67\n");
+  }
+];
+
```
jhunt: Can we add a "valid" test for something like 10.0.0.1 and 255.0.0.4?

dennisjbell: Yup, will do

dennisjbell: Ready for your re-review...

dennisjbell: Added bounds checks

Issue #223, opened 2018-02-08
Use `--no-redact` for interactive use; redact in CI/CD


jhunt:
It's really annoying to have see "<redacted>" everywhere in the confirmation of `genesis deploy` from the command line.  In very few cases do I not care about the values I'm changing, just the keys.  It's great for catching other properties changing that I didn't expect, but if its a precise change ("fix typo in base URL", for instance), I'd like to see the actual change.

CI/CD is a different story (I think) where we need to redact.  Perhaps make it a `pipeline.redact` setting in ci.yml?

dennisjbell:
Clarification:  We already have --no-redact option for genesis deploy, are you proposing to make it the default (and use --redact if you want things redacted, which I agree is almost never).

From a security perspective, I agree that we don't want the changes in the public logs, but by the same notion as above, when do we not want to see what actually changed (for confirmation the changes went out, for debugging a deployment if it failed, etc)

jhunt:
I think interactive mode should be non-redacted by default.  No one will ever use `--redact` but it's there anyway (or should be).

Pipelines should be configurable with respect to redaction - if I know who is connecting to my concourse and I trust them all with the creds, I'm cool with not redacting the build logs.  If, however, I open up the concourse pipelines to a wider audience, I may want to redact and just use the git refs for changes.  In either case, it should be a new `pipeline.secure` or `pipeline.redact` top-level configuration.

jhunt:
WIP on origin/noredact, with final commit verbiage and release notes.  Awaiting local test run to verify and then I'll submit PR.

PR #230, opened 2018-02-21
Intelligent Manifest redaction

jhunt:
The `genesis deploy` action now runs unredacted when interactive (with a
controlling terminal), and always redacted in CI/CD.


jhunt:
Fixes #223

Issue #228, opened 2018-02-21
Better Debugging


jhunt:
When Genesis runs, it would be nice if we could give it a `GENESIS_DEBUG` env var to enable debugging, and have useful output from all steps.

This probably means moving some of the `qx()` stuff over to the exec* functions, using the mkdir* helpers, etc.

jhunt:
Implemented!

Issue #227, opened 2018-02-20
Better BOSH Targeting


jhunt:
As an operator, I would like to spend less time configuring Genesis CI/CD pipelines, and minimize the amount of duplicated data in said configuration.

To achieve this, we believe the following should suffice:

1. Genesis should start tracking metadata about each deployment it has dealt with in the Vault, under `secret/genesis`.  For example, BOSH director deployments will want to provide their path into the Vault (to handle `params.vault` customizations), as well as the URL of the director:

```
$ safe get secret/genesis/some-env-bosh
---
url: https://10.0.0.4:25555
base: secret/some/env/bosh
```

This will allow other deployments to reference this director, and its secrets, as needed.  In effect, this is an extension of the core BOSH "links" concept, using Vault as a configuration server, but operating across BOSH deployments.

`genesis deploy` will be responsible for updating the entry under `secret/genesis` after every successful deployment, whether by interactive operator use, or by concourse pipeline.

Each kit will be responsible for populating other keys, besides `base` (which Genesis will get from `params.vault`, defaulting to the normal default).  These values will be placed in the new manifest top-level `genesis` key:

```
genesis:
  url: (( concat "https://" params.static_ip ":25555" ))
```

This will be imported into Vault via `safe import`, so the keys under `genesis` should only be a single level deep (no nesting!).  If the kit supplies a `genesis.base` key, it will be overwritten by genesis.

2. Environment deployments will continue to specify their BOSH director via `params.bosh`, which will be interpreted as a tertiary path component in `secret/genesis/$env-bosh`.  For example:

```
params:
  bosh: us-east-1-prod
```

would be interpreted as `secret/genesis/us-east-1-prod-bosh` for lookups.

3. Add a new command called `genesis login` that will do the required `bosh` CLI calls to set up an alias, using the correct parameters:

```
bosh -e $URL logout
bosh -e $URL --ca-cert "$CA" alias-env $ENV
bosh -e $ENV env
bosh -e $ENV login
```

This will make life easier on operators, who are tired of dealing with BOSH's quirks on the command-line.

4. For CI/CD use cases, the ci.yml will pick up a new top-level param called `bosh_client` (or `bosh-client`, whatever is consistenter) that will identify the name of the UAA client to use when talking to BOSH.  It will default to `concourse`.

[...]

PR #240, opened 2018-03-16
Pre genesis metadata fixes

### LINKED ISSUE #237: Should compiled kits contain the ci directory
When we compile kits, we currently include the ci directory that is used to test the kit.  Is this necessary?  Do we use it from the compiled kit when testing, or do we just test the repo?

### LINKED ISSUE #238: `genesis deploy` and `genesis manifest` should check for env file existence
Did this in the lab today:

```
genesis manifest dbell-test
Could not find `params.bosh` or `params.env` in environment file
```

The file was named `dbell-testing`.

We should probably verify that the tail file in the ancestry exists, and emit a better message.

dennisjbell:
This are issues that were fixed en-masse in support of or discovered during the move to use genesis metadata in vault to better support BOSH discovery for pipelines (#227). They include:

* Fix default redaction behaviour on manifest and deploy (redacts on pipeline or redirection, doesn't redact from a controlling terminal) .  Still can be manually specified to redact or not.
* Added --dry-run on deploy to match BOSH's deploy --dry-run behaviour.
* Improved execution of subprocesses (phase 1 of issue #228 )
* Improved `has_key` and `get_key` (1d8e0ca)
* -D option is now less noisy: Sets GENESIS_DEBUG instead of DEBUG so you only get Genesis's debug output, not including all the subprocesses (ie spruce).  Setting DEBUG env var prior to calling still results in the full debug effect.
* Ensures the specified env file exists for manifest and deploy actions (Fixes #238)
* Un-deprecates the subkit usage until 'features' are officially released and supported
* Removed ci directory from compiled kits (Fixes #237 )

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -156,7 +156,7 @@ sub explain(@) {
 }
 
 sub debug(@) {
-	return unless envset "DEBUG";
+	return unless envset "GENESIS_DEBUG";
 	print STDERR "DEBUG> ";
 	my $colorize = $ENV{NOCOLOR};
 	$ENV{NOCOLOR} = "true" if (! -t STDERR);
@@ -205,10 +205,10 @@ sub controlling_terminal {
 #
 # The latter is recommended as it properly encapsulates/tokenizes the 
 # arguments to prevent accidental expansion or splitting due to quoting
-# and spaces.
+# and spaces.  If using it, remember to quote all variable references.
 #
 # Use this where you would normally use backticks or qx() calls.  It also
-# supports debug output when DEBUG env is set.
+# supports debug output when GENESIS_DEBUG env is set.
 #
 # If called in a scalar context, it will return true if success, false
 # otherwise.  In a list context, it will return the output and the exit code
@@ -1665,7 +1665,7 @@ sub write_stemcell_data {
 	my $stemcells = "{}";
 	if ($ENV{GENESIS_INDEX} ne "no") {
 		my $flags = "-ks";
-		$flags += " -v" if envset 'DEBUG';
+		$flags += " -v" if envset 'GENESIS_DEBUG';
 		$stemcells = `curl $flags $ENV{GENESIS_INDEX}/v1/stemcell/latest | jq '. | map({"key": .name, "value": { "sha1": .sha1, "url": .url}}) | from_entries' -Mc`;
 		die "Unable to contact the genesis index to retrieve stemcell data. Cannot continue." unless ($? == 0);
 	}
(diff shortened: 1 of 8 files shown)


bin/genesis
```diff
@@ -195,20 +195,41 @@ sub controlling_terminal {
 	return -t STDOUT;
 }
 
+# Execute the requested command.  You can operate this in two modes:
+#   1) Pass the command as a string will all the arguments in it.
```
dennisjbell: will --> with

bin/genesis
```diff
@@ -195,20 +195,41 @@ sub controlling_terminal {
...
+#
+# If called in a scalar context, it will return true if success, false
+# otherwise.  In a list context, it will return the output and the exit code
 sub execute {
-	my ($prog) = @_;
-	debug "executing `#C{$prog} 2>&1`";
-	my $out = qx($prog 2>&1);
-	my $rc = $?;
+	my ($prog,@args) = @_;
+	my $shell = $ENV{SHELL} || '/bin/bash';
```
dennisjbell: Just use /bin/bash

bin/genesis
```diff
@@ -195,20 +195,41 @@ sub controlling_terminal {
...
+	debug "#M{Executing:} `#C{$prog 2>&1}`";
+	if (@args) {
+		unshift @args, $ENV{SHELL};
```
dennisjbell: Use '/bin/bash' explicitly

bin/genesis
```diff
@@ -195,20 +195,41 @@ sub controlling_terminal {
 	return -t STDOUT;
 }
 
+# Execute the requested command.  You can operate this in two modes:
+#   1) Pass the command as a string will all the arguments in it.
+#   2) Pass the command as a template, and the template args as an array
+#
```
dennisjbell: Examples please

dennisjbell: and for the scalar/list call context

bin/genesis
```diff
@@ -935,43 +934,61 @@ sub get_env_file {
...
-			return $default unless @{$result}[0];
-			$result = @{$result}[0];
-		} else {
-			return $default unless exists $result->{$_};
-			$result = $result->{$_};
-		}
-	}
-	return defined $result ? $result : $default;
+	my ($out,$rc) = execute(
+		'f=$1; shift; spruce merge --skip-eval $@ | spruce json | jq -Mc $f',
```
dennisjbell: Quote all variables

bin/genesis
```diff
@@ -5327,6 +5342,7 @@ sub {
 		dry-run
 	/);
 	usage(1) if @_ != 1;
+	my $env = get_env(shift);;
```
dennisjbell: ;; -> ;

PR #242, opened 2018-03-17
set HTTPS_PROXY=$BOSH_ALL_PROXY if available; unset HTTPS_PROXY for 'bosh' CLI calls

drnic:
This feature will require https://github.com/geofffranks/spruce/pull/253 to be merged

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -22,6 +22,10 @@ our $VERSION = "(development)";
 our $GITHUB_URL = "https://github.com/starkandwayne/genesis";
 our $USER_AGENT_STRING = "genesis/$VERSION";
 
+if ($ENV{BOSH_ALL_PROXY}) {
+	$ENV{HTTPS_PROXY}=$ENV{BOSH_ALL_PROXY};
+}
+
 sub DumpJSON {
 	my ($file, $data) = @_;
 	open my $fh, ">", $file or die "Unable to write to $file: $!\n";
@@ -1550,6 +1554,16 @@ sub detect_bosh_version {
 	}
 }
 
+sub system_bosh {
+  local $ENV{HTTPS_PROXY} = ''; # bosh dislikes this env var
+  return system_execute(join(" ", $BOSH, @_));
+}
+
+sub qx_bosh {
+  local $ENV{HTTPS_PROXY} = ''; # bosh dislikes this env var
+  return qx(join(" ", $BOSH, @_));
+}
+
 sub is_create_env {
 	my ($env) = @_;
 	# convert env to file for env if not already done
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -22,6 +22,10 @@ our $VERSION = "(development)";
 our $GITHUB_URL = "https://github.com/starkandwayne/genesis";
 our $USER_AGENT_STRING = "genesis/$VERSION";
 
+if ($ENV{BOSH_ALL_PROXY} ne '') {
```
jhunt: If BOSH_ALL_PROXY isn't set, this throws an uninitialized value, since you are comparing `undef` to the empty string.  You can simplify by just using `$ENV{BOSH_ALL_PROXY}` as your conditional -- in Perl, the empty string, false, 0, and the undefined value all evaluate to false in truthy context.

bin/genesis
```diff
@@ -220,6 +224,7 @@ sub system_execute {
 	return 1;
 }
 
+
```
jhunt: Can you remove the extraneous whitespace?

bin/genesis
```diff
@@ -1550,6 +1555,30 @@ sub detect_bosh_version {
 	}
 }
 
+sub system_bosh {
+	my ($cmd) = join(" ", ($BOSH, @_));
```
jhunt: Perl automatically flattens lists, so `join(' ', ($scalar, @list)` is identical to `join(' ', $scalar, @list)`

You can remove the superfluous parentheses.

bin/genesis
```diff
@@ -1550,6 +1555,30 @@ sub detect_bosh_version {
...
+	if ($ENV{BOSH_ALL_PROXY} ne '') {
+		$ENV{HTTPS_PROXY}='';
+	}
+	my @result = system_execute($cmd);
+	if ($ENV{BOSH_ALL_PROXY} ne '') {
+		$ENV{HTTPS_PROXY}=$ENV{BOSH_ALL_PROXY};
+	}
+	return @result;
+}
```
jhunt: This pattern of chage-then-restore is better handled using local scope, and the Perl `local` keyword.

If you declare a variable using `local` (as opposed to `my` or `our`), the binding will only persist until the stack is unwound at function return.

This function then becomes something like:

```
sub system_bosh {
  local $ENV{HTTP_PROXY} = ''; # bosh dislikes this env var
  return system_execute(join(" ", $BOSH, @_);
}
```

drnic: Awesome. Much nicer.

bin/genesis
drnic: Lovely.

PR #247, opened 2018-03-19
Revamp the way external processes are run.

jhunt:
The new `run()` function handles all of the different use cases we have,
and attempts to be seamless and flexible.  It has wrappers in `safe` and
`bosh` that handle configuration (like interactivity) and it lends
itself to use in `run(...) or die ...` idioms.


jhunt:
This is absolutely terrible.  Whoever did this should be ashamed of him- or herself.

PR #248, opened 2018-03-20
Genesis::ShellExec refactor

dennisjbell:
This PR introduces the Genesis::ShellExec package, a unifying package meant to replace all the various methods used to execute shell commands, such as backticks, qx() and shell, as well as the derived execute, system_execute, system_bosh and qx_bosh subroutines.

It promotes tokenization of variables to prevent inadvertent expansion, and logs all the calls to `STDERR` when `-D/--debug` option is turned on (or DEBUG environment variable is true)

You can operate this in two modes:
1. Pass the command as a string with all the arguments in it.
```
execute("safe read auth/approle/role/$role_name/role-id | spruce json | jq -r .role_id | safe paste $role_p $role_k");
```

2. Pass the command as a template, and the template args as an array
```
execute('safe read "$1" | spruce json | jq -r .role_id | safe paste "$2" "$3"',
    "auth/approle/role/$role_name/role-id",
    $role_p,
    $role_k
);
```

The latter is recommended as it properly encapsulates/tokenizes the arguments to prevent accidental expansion or splitting due to quoting and spaces.  If using it, remember to quote all variable references.

You can also pass a hash reference as the first argument to supply options to change the behaviour of the execution.  The following options are supported:

* `interactive: boolean` - If true, run the command interactively on a controlling terminal.  This uses the perl `system` command, so capturing output cannot be done.  Returns exit code.
* `passfail: boolean` - If true, returns true if exit code is 0, false otherwise.  Can work in conjunction with interactive.
* `onfailure: string` - Print the string on error and exit program.  If not running interactively, also prints the output received.
* ` env: hash` - Sets the env variables specified as keys in the hash to the corresponding values before executing the command.
* `shell: string` - By default, will use /bin/bash, but if the command needs a different shell, specify it with this option.
 * `stderr: string` - By default, stderr is redirected to stdout (&1), but you can use this option to redirect it to a file for later use.  Cannot use with interactive mode.

Finally, if you are not running in interactive or passfail mode, you can call this in either scalar or list context.  If calling in list context, the output and exit code are returned in a list, otherwise just the output.

Scalar context:
```
my $out = execute('grep "$1" "$@"', $pattern, @files);
my $rc = $? >> 8;
```
List context:
```
my ($out,$rc) = execute('spruce json "$1" | jq -r "$2"', $_, $filter);
```

It also supports several common use cases with convenience functions:
* `interact`
  - sets the mode to interactive and passfail
* `bosh`
  - sets the environment for bosh and injects the correct bosh executable
* `bosh_interact`
  - same as bosh, but using interactive and passfail modes
* `check `
  - sets passfail mode, no output returned.
* `do_or_die`
  - first argument is printed along with any output received, then exits non-zero (dies) if remaining argument fails execution
* `get`
  - returns just output, regardless of context
* `getlines`
  - returns output as an array of lines

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
 	$ENV{HTTPS_PROXY}=$ENV{BOSH_ALL_PROXY};
 }
 
+package Genesis::ShellExec;
+
+# This package provides a unified method of executing shell commands, and some
+# convenience methods to run them in common ways.
+#
+# You can operate this in two modes:
+#
+#   1) Pass the command as a string with all the arguments in it.
+#      eg: execute("safe read auth/approle/role/$role_name/role-id | spruce json | jq -r .role_id | safe paste $role_p $role_k");
+#
+#   2) Pass the command as a template, and the template args as an array
+#      eg: execute('safe read "$1" | spruce json | jq -r .role_id | safe paste "$2" "$3"',
+#                  "auth/approle/role/$role_name/role-id",
+#                  $role_p,
+#                  $role_k
+#          );
+#
+# The latter is recommended as it properly encapsulates/tokenizes the
+# arguments to prevent accidental expansion or splitting due to quoting
+# and spaces.  If using it, remember to quote all variable references.
+#
+# You can also pass a hash reference as the first argument to supply options to
+# change the behaviour of the execution.  The following options are supported:
+#
+#   interactive: boolean - If true, run the command interactively on a
+#                controlling terminal.  This uses the perl `system` command, so
+#                capturing output cannot be done.  Returns exit code.
+#
+#   passfail:    boolean - If true, returns true if exit code is 0, false
+#                otherwise.  Can work in conjunction with interactive.
+#
+#   onfailure:   string - Print the string on error and exit program.  If not
+#                running interactively, also prints the output received.
+#
+#   env:         hash - Sets the env variables specified as keys in the hash to
+#                the corresponding values before executing the command.
+#
+#   shell:       string - By default, will use /bin/bash, but if the command
+#                needs a different shell, specify it with this option.
+#
+#   stderr:      string - By default, stderr is redirected to stdout (&1), but
+#                you can use this option to redirect it to a file for later
+#                use.  Cannot use with interactive mode.
+#
+# Finally, if you are not running in interactive or passfail mode, you can call
+# this in either scalar or list context.  If calling in list context, the output
+# and exit code are returned in a list, otherwise just the output.
+#
+# Scalar context:
+#   my $out = execute('grep "$1" "$@"', $pattern, @files);
+#   my $rc = $? >> 8;
+#
+# List context:
+#   my ($out,$rc) = execute('spruce json "$1" | jq -r "$2"', $_, $filter);
+#
+# -----------------------------------------------------------------------------
+# Convenience commands:
+#
+# interact - sets the mode to interactive and passfail
+#
+# bosh - sets the environment for bosh and injects the correct bosh executable
+#
+# bosh_interact - same as bosh, but using interactive and passfail modes
+#
+# check - sets passfail mode, no output returned.
+#
+# do_or_die: first argument is printed along with any output received, then
+#            exits non-zero (dies) if remaining argument fails execution
+#
+# get - returns just output, regardless of context
+#
+# getlines - returns output as an array of lines
+sub execute {
+	my (@args) = @_;
+	my %opts = %{((ref($args[0]) eq 'HASH') ? shift @args: {})};
+	my $prog = shift @args;
+	if ($prog !~ /\$\{?[\@0-9]/ && scalar(@args) > 0) {
+		$prog .= ' "$@"'; # old style of passing in args as array, need to wrap for shell call
+	}
+
+	local %ENV = %ENV; # To get local scope for duration of this call
+	for (keys %{$opts{env} || {}}) {
+		$ENV{$_} = $opts{env}{$_};
+		main::debug("#M{Setting: }#B{$_}='#C{$ENV{$_}}'");
+	}
+	my $shell = $opts{shell} || '/bin/bash';
+	$prog .= ($opts{stderr} ? " 2>$opts{stderr}" : ' 2>&1') unless ($opts{interactive});
+	main::debug("#M{Executing:} `#C{$prog}`%s", ($opts{interactive} ? " #Y{(interacively)}" : ''));
+	if (@args) {
+		unshift @args, $shell;
+		main::debug("#m{ - with arguments:}");
+		main::debug("#m{%4s:} '#c{%s}'", $_, $args[$_]) for (1..$#args);
+	}
+
+	my @cmd = ($shell, "-c", $prog, @args);
+	my $out;
+	if ($opts{interactive}) {
+		system @cmd;
+	} else {
+		open my $pipe, "-|", @cmd;
+		$out = do { local $/; <$pipe> };
+		close $pipe;
+	}
+	my $rc = $? >>8;
+	if ($rc) {
+		die("#R{$opts{onfailure}}%s\n", defined($out) ? ":\n$out" :'') if $opts{onfailure};
+		main::debug("#R{==== ERROR: $rc}");
+		main::debug("$out\n#R{==== END}") if defined($out);
+	} else {
+		main::debug("#g{Command executed successfully.}");
+	}
+	return
+		$opts{passfail}    ? $rc == 0 :
+		$opts{interactive} ? $rc :
+		wantarray          ? ($out,$rc) :
+		$out;
+}
+
+sub interact {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	$opts->{interactive} = 1;
+	$opts->{passfail} = 1;
+	return execute($opts,@args);
+}
+
+sub check {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	$opts->{interactive} = 0;
+	$opts->{passfail} = 1;
+	return execute($opts,@args);
+}
+
+sub bosh {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	$opts->{env} ||= {};
+	# Clear out the BOSH env vars unless we're under Concourse
+	unless ($ENV{BUILD_PIPELINE_NAME}) {
+		$opts->{env}{HTTPS_PROXY} = ''; # bosh dislikes this env var
+		$opts->{env}{https_proxy} = ''; # bosh dislikes this env var
+		$opts->{env}{BOSH_ENVIRONMENT}   ||= ''; # ensure using ~/.bosh/config via alias
+		$opts->{env}{BOSH_CA_CERT}       ||= '';
+		$opts->{env}{BOSH_CLIENT}        ||= '';
+		$opts->{env}{BOSH_CLIENT_SECRET} ||= '';
+	}
+	if ($args[0] =~ /\$\{?[\@0-9]/ && scalar(@args) > 1) {
+		$args[0] = main::bosh_cmd() ." $args[0]";
+	} else {
+		unshift @args, main::bosh_cmd()
+	}
+	return wantarray ? (execute($opts,@args)) : execute($opts,@args);
+}
+
+sub bosh_interact {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	$opts->{interactive} = 1;
+	$opts->{passfail} = 1;
+	return bosh($opts,@args);
+}
+
+sub do_or_die {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	delete @{$opts}{qw/interative passfail/};
+	$opts->{onfailure} = shift @args;
+	execute($opts,@args);
+	return undef;
+}
+
+sub get {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	delete @{$opts}{qw/interative passfail/};
+	my $out = execute($opts,@args);
+	return $out
+}
+
+sub getlines {
+	my $out = get(@_);
+	return ($? >> 0) ? () : split $/, $out;
+}
+package main;
+
 sub DumpJSON {
 	my ($file, $data) = @_;
 	open my $fh, ">", $file or die "Unable to write to $file: $!\n";
@@ -35,7 +221,7 @@ sub DumpJSON {
 
 sub LoadFile {
 	my ($file) = @_;
-	decode_json(qx(spruce json $file));
+	decode_json(Genesis::ShellExec::get('spruce json "$1"', $file));
 }
 
 sub Load {
@@ -73,7 +259,9 @@ sub envdefault {
 sub valid_envs {
 	my (%envs, $json);
 	foreach (glob "*.yml") {
-		eval {$json = decode_json(qx{spruce json "$_" 2>/dev/null});};
+		eval {
+			$json = decode_json(Genesis::ShellExec::get({stderr=>"/dev/null"}, 'spruce json "$1"', $_))
+		};
 		next if $@;
 		next unless $json && $json->{params} && $json->{params}{env};
 		(my $label= $_) =~ s/\.yml$//;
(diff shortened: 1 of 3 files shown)


bin/genesis
```diff
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
...
+		$opts->{env}{BOSH_CA_CERT}       ||= '';
+		$opts->{env}{BOSH_CLIENT}        ||= '';
+		$opts->{env}{BOSH_CLIENT_SECRET} ||= '';
+	}
+	if ($args[0] =~ /\$\{?[\@0-9]/ && scalar(@args) > 1) {
+		$args[0] = main::bosh_cmd() ." $args[0]";
+	} else {
+		unshift @args, main::bosh_cmd()
+	}
+	return wantarray ? (execute($opts,@args)) : execute($opts,@args);
```
jhunt: Let the callers context flow through here.

bin/genesis
```diff
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
...
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	$opts->{interactive} = 1;
+	$opts->{passfail} = 1;
+	return bosh($opts,@args);
+}
+
+sub do_or_die {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	delete @{$opts}{qw/interative passfail/};
```
jhunt: typo on "interative" (missing a C)

(or don't delete the interactive flag)

bin/genesis
```diff
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
...
+	$opts->{onfailure} = shift @args;
+	execute($opts,@args);
+	return undef;
+}
+
+sub get {
+	my @args = @_;
+	my $opts = (ref($args[0]) eq 'HASH') ? shift @args : {};
+	delete @{$opts}{qw/interative passfail/};
```
jhunt: interative again.  this one definitely needs to be "interactive"

bin/genesis
```diff
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
...
+	my $out = execute($opts,@args);
+	return $out
```
jhunt: If we modify execute() to return ($output, $rc) in list context, and $rc == 0 in scalar context, we can do this:

```
(my $out, undef) = execute($opts, @args);
return $out;
```

bin/genesis
```diff
@@ -26,6 +26,192 @@ if ($ENV{BOSH_ALL_PROXY}) {
...
+}
+
+sub getlines {
+	my $out = get(@_);
+	return ($? >> 0) ? () : split $/, $out;
```
jhunt: This is not how you want to pull the rc.  You probably meant `>> 8`

bin/genesis
```diff
@@ -428,8 +553,7 @@ sub run_feature_hook {
 	$version = "latest" unless defined $version;
 	chmod(0755,$hook) unless -x $hook;
 
-	my $cmd = "$hook " . join(" ", map { "'$_'" } @features);
-	my $out = qx($cmd);
+	my $out = Genesis::ShellExec::get('cmd="$1"; shift; "$cmd" "$@"', $hook, @features);
```
jhunt: Why not

```
my $out = Genesis::ShellExec::I::Love::Long::Module::Names::get($hook, @features)
```
?

bin/genesis
```diff
@@ -449,10 +573,14 @@ sub run_param_hook {
...
-	$ENV{GENESIS_ENVIRONMENT_NAME} = $env_name;
-	$ENV{GENESIS_VAULT_PREFIX} = $vault_prefix;
-	system($cmd);
+	Genesis::ShellExec::interact(
+		{env => {
+			GENESIS_ENVIRONMENT_NAME => $env_name,
+			GENESIS_VAULT_PREFIX => $vault_prefix
+		}},
+		'dir="$1"; shift 1; "'.$hook.'" "${dir}/in" "${dir}/out" "$@"',
+		$dir, @features
```
jhunt: same here, use the arrayified version of get() / execute()

bin/genesis
```diff
@@ -617,7 +745,7 @@ sub kit_yaml_files_in {
 		-x $blueprint_script_path || chmod_or_fail 0755,$blueprint_script_path;
 		my $helper = blueprint_script_helper($kit,$version,$env,@features);
 		pushd($dir);
-		my $result = qx(bash -c 'source $helper; hooks/blueprint');
+		my $result = Genesis::ShellExec::get('source "$1"; hooks/blueprint', $helper);
```
jhunt: Use a get() with an onfailure.

bin/genesis
```diff
@@ -908,8 +1036,8 @@ sub get_kit {
 	unless (exists $KIT_CACHE_DIR{"$kit/$version"}) {
 		my $dir = workdir('kit');
 		die "Kit $kit/$version not found!\n" unless -e ".genesis/kits/$kit-$version.tar.gz";
-		qx(tar -xz -C $dir --strip-components 1 -f .genesis/kits/$kit-$version.tar.gz);
-		$? == 0 or die;
+		Genesis::ShellExec::check('tar -xz -C "$1" --strip-components 1 -f ".genesis/kits/${2}-${3}.tar.gz"', $dir, $kit, $version)
+			or die "Could not read kit file";
```
jhunt: Make this a do_or_die?

bin/genesis
```diff
@@ -1013,11 +1141,11 @@ sub get_key {
 	return $default unless has_key($file, $key);
 	my $filter = build_jq_filter($key);
 
-	my ($out,$rc) = execute(
+	my $out = Genesis::ShellExec::execute(
```
jhunt: This is identical to get(); use get() instead.

bin/genesis
```diff
@@ -1044,8 +1172,7 @@ sub dereference_params {
 }
 
 sub safe_path_exists {
-	system "safe exists $_[0]";
-	return ( $? eq 0);
+	return Genesis::ShellExec::check('safe exists "$1"', $_[0]);
```
jhunt: Use the arrayified version of execute()

bin/genesis
```diff
@@ -1608,44 +1732,29 @@ sub is_valid_uri {
...
-		my $version = qx($boshcmd -v 2>&1 | grep version | head -n1);
-		if ($version) {
-			chomp($version);
-			if ($version =~ s/version (\S+?)-.*/$1/){
-				if (!new_enough("bosh2", $version, $BOSH2_MIN_VERSION)) {
-					error "BOSH (v2) CLI v${version} is installed, but Genesis requires #R{at least $BOSH2_MIN_VERSION} - upgrade your BOSH CLI, via #B{https://github.com/cloudfoundry/bosh-cli/releases}";
-				}
+	my $best = "0.0.0";
+	foreach my $boshcmd (qw(bosh2 boshv2 bosh)) {
+		chomp(my $version = Genesis::ShellExec::execute("$boshcmd -v 2>&1 | grep version | head -n1"));
```
jhunt: Change this to a get.

jhunt: I think we can have `get` chomp its output.

bin/genesis
```diff
@@ -1667,27 +1776,29 @@ sub bosh_deploy {
...
+	Genesis::ShellExec::bosh_interact(
+		{onfailure => "Could not create BOSH alias for '$_[0]'"},
+		'alias-env "$1"', $_[0]
+	);
 }
 
 sub bosh_run_errand {
 	my ($target, $deployment, $errand) = @_;
-	system_bosh("-n", "-e", $target, "-d", $deployment, "run-errand", $errand) or exit 1;
+	Genesis::ShellExec::bosh_interact("-n", "-e", $target, "-d", $deployment, "run-errand", $errand) or exit 1;
```
jhunt: Specify an error message

bin/genesis
```diff
@@ -1667,27 +1776,29 @@ sub bosh_deploy {
...
 }
 
 sub bosh_upload_stemcell {
 	my ($name, $version, $sha1, $url) = @_;
-	system_bosh("-n", "upload-stemcell", "--name", $name, "--version", $version, "--sha1", $sha1, $url) or exit 1;
+	Genesis::ShellExec::bosh_interact("-n", "upload-stemcell", "--name", $name, "--version", $version, "--sha1", $sha1, $url) or exit 1;
```
jhunt: Specify an error message

bin/genesis
```diff
@@ -1858,27 +1971,17 @@ sub deploy_manifest {
...
-	}
-	if (envset('GENESIS_DEBUG')) {
-		$flags .= " -v";
-	}
+	my @flags = ("-X", $method);
+	push @flags, "-H", "$_: $headers->{$_}" for (keys %$headers);
+	push @flags, "-d", $data if $data;
+	push @flags, "-k"  if ($skip_verify);
+	push @flags, "-u", $creds if ($creds);
+	push @flags, "-v"  if (envset('GENESIS_DEBUG'));
```
jhunt: Line up the ifs.

bin/genesis
```diff
@@ -3444,7 +3554,7 @@ sub check_prereqs {
 	my $version;
 
 	# check that we has a spruce
-	$version = qx(spruce -v 2>/dev/null);
+	$version = Genesis::ShellExec::get {stderr => '/dev/null'}, 'spruce -v';
```
jhunt: Parenthesize the args.

bin/genesis
```diff
@@ -3457,7 +3567,7 @@ sub check_prereqs {
 	}
 
 	# check that we has a safe
-	$version = qx(safe -v 2>&1 >/dev/null);
+	$version = Genesis::ShellExec::get {stderr => "&1 >/dev/null"}, 'safe -v'; # kinda hacky way of getting stdout to /dev/null and stderr to stdout
```
jhunt: Parenthesize the args.

bin/genesis
```diff
@@ -3470,7 +3580,7 @@ sub check_prereqs {
 	}
 
 	# check that we has a vault
-	$version = qx(vault -v 2>/dev/null);
+	$version = Genesis::ShellExec::get {stderr => '/dev/null'}, 'vault -v';
```
jhunt: Parenthesize the args.

bin/genesis
```diff
@@ -3567,7 +3677,11 @@ sub kit_file {
 		$dir = $KIT_CACHE_DIR{"$kit/$version"};
 	} else {
 		$dir = workdir('kit');
-		qx(tar -C $dir -xz --strip-components 1 -f .genesis/kits/$kit-$version.tar.gz $kit-$version/$relpath 2>/dev/null);
+		Genesis::ShellExec::check(
```
jhunt: Do an execute() here, since a check() makes it seem like we should be reacting.

Otherwise, let's react to the failure?

bin/genesis
```diff
@@ -4069,12 +4191,17 @@ sub active_certificates {
 sub target_vault {
 	my ($target) = @_;
 	if ($target) {
-		system(qw(safe target), $target) == 0 or exit 1;
+		Genesis::ShellExec::interact({passfail => 1}, 'safe target "$1"', $target) or exit 1;
 	} else {
-		system(qw(safe target -i)) == 0 or exit 1;
+		Genesis::ShellExec::interact({passfail => 1}, 'safe target -i') or exit 1;
```
jhunt: Ours is not to question why.

Ours is but to do_or_die()

(with interact => 1)

jhunt: Migrate the if ($target) call into the arg list of the interact.

The target is either set, or it's `-i`

bin/genesis
```diff
@@ -4069,12 +4191,17 @@ sub active_certificates {
...
 	}
-	chomp($ENV{GENESIS_TARGET_VAULT}=qx(safe target 2>&1 | grep '.*targeting .* at.*' | sed -e 's/.*targeting \\([^ ]*\\) at.*/\\1/'));
-	die "Could not set target vault" if $? >> 8;
+	chomp(
+		$ENV{GENESIS_TARGET_VAULT}=Genesis::ShellExec::get(
+			{onfailure => "Could not set target vault"} ,
+			'safe target 2>&1 | grep "$1" | sed -e "$2"',
+			'.*targeting .* at.*', 's/.*targeting \([^ ]*\) at.*/\1/'
+		)
+	);
```
jhunt: Put a FIXME comment in here.

safe now supports `safe target --json`, and once that gets released, we should totes use that.

bin/genesis
```diff
@@ -4107,8 +4234,10 @@ sub vaultify_secrets {
 	if (%$creds) {
 		explain " - auto-generating credentials (in secret/$options{prefix})...\n";
 		for (safe_commands $creds, %options) {
-			system('safe', @$_);
-			die "Failure autogenerating credentials.\n" if ($? >> 8);
+			Genesis::ShellExec::interact(
+				{onfailure => "Failure autogenerating credentials."},
+				'safe "$@"', @$_
```
jhunt: Can we let execute() append the ` "$@"` for us?

bin/genesis
```diff
@@ -4118,8 +4247,10 @@ sub vaultify_secrets {
 	if (%$certs) {
 		explain " - auto-generating certificates (in secret/$options{prefix})...\n";
 		for (cert_commands $certs, %options) {
-			system('safe', @$_);
-			die "Failure autogenerating certificates.\n" if ($? != 0);
+			Genesis::ShellExec::interact(
+				{onfailure => "Failure autogenerating certificates."},
+				'safe "$@"', @$_
+			);
```
jhunt: same here

bin/genesis
```diff
@@ -4190,18 +4321,18 @@ sub tablify {
 sub bosh_target_for {
 	my ($env) = @_;
 	if (defined($ENV{GENESIS_BOSH_ENVIRONMENT})) {
-		system_bosh("-e $ENV{GENESIS_BOSH_ENVIRONMENT} env") and
+		Genesis::ShellExec::bosh_interact('-e "$1" env', $ENV{GENESIS_BOSH_ENVIRONMENT}) and
 			return $ENV{GENESIS_BOSH_ENVIRONMENT};
 		die "No such host: $ENV{GENESIS_BOSH_ENVIRONMENT} from GENESIS_BOSH_ENVIRONMENT environment variable for the BOSH Director.\n";
```
jhunt: Lose the `-e` and set `BOSH_ENVIRONMENT` in the env hash.

Also, this should be a do_or_die, followed by a return of the env var

bin/genesis
```diff
@@ -4190,18 +4321,18 @@ sub tablify {
...
 	}
 	my $bosh = get_key($env, 'params.bosh');
 	if (defined($bosh)) {
-		system_bosh("-e $bosh env") and return $bosh;
+		Genesis::ShellExec::bosh_interact('-e "$1" env', $bosh) and return $bosh;
```
jhunt: Use the `env` hash here, and also make it a do_or_die followed by a return.

bin/genesis
```diff
@@ -4190,18 +4321,18 @@ sub tablify {
...
 		die "No such host: $bosh for the BOSH Director which you configured in params.bosh.\n";
 	 }
 	$bosh = get_key($env, 'params.env');
 	if (defined($bosh)) {
-		system_bosh("-e $bosh env") and return $bosh;
+		Genesis::ShellExec::bosh_interact('-e "$1" env', $bosh) and return $bosh;
 		die "No such host: $bosh for the BOSH Director which you configured in params.env. You can specify your BOSH Director alias name in params.bosh.\n";
```
jhunt: same

bin/genesis
```diff
@@ -4426,9 +4557,9 @@ sub {
 	debug "generating a new Genesis repo, named $name";
 
 	debug "checking git config so can git commit later";
-	execute 'git config user.name'
+	Genesis::ShellExec::check 'git config user.name'
 		or die 'Please setup git - git config --global user.name "Your Name" -';
```
jhunt: Should this be a do_or_die?

bin/genesis
```diff
@@ -4426,9 +4557,9 @@ sub {
...
-	execute 'git config user.email'
+	Genesis::ShellExec::check 'git config user.email'
 		or die 'Please setup git - git config --global user.email your@email.com -';
```
jhunt: Should this be a do_or_die?

bin/genesis
```diff
@@ -4561,11 +4692,11 @@ EOF
...
 
-	execute 'git init'
+	Genesis::ShellExec::check 'git init'
 		or die "Failed to initialize a git repository in $root/\n";
-	execute 'git add .'
+	Genesis::ShellExec::check 'git add .'
 		or die "Failed to stage files to git, for initial commit, in $root/\n";
-	execute 'git commit -m "Initial Genesis Repo"'
+	Genesis::ShellExec::check 'git commit -m "Initial Genesis Repo"'
 		or die "Failed to commit initial Genesis repository in $root/\n";
```
jhunt: Should this be a do_or_die?

bin/genesis
```diff
@@ -5653,8 +5803,8 @@ EOF
 		print STDERR csprintf("\n#R{Warning:} Using development version of Genesis -- cannot specify minimum Genesis version\n");
 	}
 
-	chomp(my $user = qx'git config user.name');
-	chomp(my $email = qx'git config user.email');
+	chomp(my $user = Genesis::ShellExec::get('git config user.name'));
+	chomp(my $email = Genesis::ShellExec::get('git config user.email'));
```
jhunt: It definitely feels like we should chomp() in get()

bin/genesis
```diff
@@ -5807,20 +5957,22 @@ sub {
...
-		my $changes = qx(git status --porcelain);
+		my $changes = Genesis::ShellExec::check('git status --porcelain');
 		if ($? == 0 && $changes) {
 			error "\n#R{ERROR}: Unstaged / uncommited changes found in working directory.\n".
 			      "Please either #C{stash} or #C{commit} those changes before compiling your kit.\n";
 			exit 1;
 		}
 	}
-	qx(tar -czf $stem.tar.gz -C $temp $stem/);
+	Genesis::ShellExec::check('tar -czf "${2}.tar.gz" -C "$1" "$2/"', $temp, $stem);
```
jhunt: It seems like all these check() calls should be reacted to.  Might I interest you in `Genesis::ShellExec::Some::More::Modules::For::Fun::run()`?

PR #249, opened 2018-03-20
'genesis new' - choose env from known bosh aliases

drnic:
Previously the user had to type a label for their genesis/bosh env 
`genesis new some-env` and the subsequent `genesis deploy` would discover
if `some-env` was actually a BOSH environment alias or not.

This PR ensures that the user already has `bosh alias-env` for each environment
by asking the user to choose their `genesis new` from a menu of `bosh envs`.

Output looks like:

```
$ genesis new

Choose a BOSH/Genesis environment:
  1) useast2-prod (10.10.1.4)
  2) useast2-lite (10.10.1.5)

Select choice > useast2-lite (10.10.1.5)
Generating new environment useast2-lite...
```

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -4729,7 +4729,7 @@ EOF
 
 command("new", <<EOF,
 genesis v$VERSION
-USAGE: genesis new [--vault target] env-name[.yml]
+USAGE: genesis new [--vault target]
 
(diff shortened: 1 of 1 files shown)


drnic:
I'll close this. I believe I worked around it ok.

PR #278, opened 2018-05-10
Set BOSH director environment variables for hooks

### LINKED ISSUE #277: Automatically Target BOSH in hook scripts
As it stands, we are relying on the hook scripts to use the `bosh` CLI properly and we give them an alias name.  This is sub-optimal.

I'd much rather hook scripts operate connected _to_ BOSH already, via the environment variables:

- **BOSH_ENVIRONMENT** - Director URL
- **BOSH_CA_CERT** - Contents of or path to CA that signed the director's certificate
- **BOSH_CLIENT_ID** - UAA Client ID for authentication
- **BOSH_CLIENT_SECRET** - UAA Client Secret for authentication

We can support this today without ripping all the alias-centric guts out of Genesis itself.  The hooks are more durable and long-lived than the `genesis` script anyway.

The gist of the idea is this:

```
spruce json ~/.bosh/config  | jq -r '.environments[] | select(.alias == "lab")'
{
  "alias": "lab",
  "ca_cert": "-----BEGIN CERTIFICATE-----\n... etc ...\n-----END CERTIFICATE-----\n\n",
  "password": "sekrit",
  "url": "https://10.0.0.4",
  "username": "admin"
}
```

In general, we conflate users/clients and make sure that the admin user/client have the same password/secret, so this _should_ work in the short-term.

The primary benefit is that it allows hook scripts to just run `bosh ...` if they need access to the _deploying_ BOSH director (for example, to look up dynamic IPs, or run errands)

jhunt:
When any hook (except prereqs and new) runs, it now gets the following
four environment variables, to make dealing with BOSH seamless:

  - BOSH_ENVIRONMENT   - Director URL
  - BOSH_CA_CERT       - Contents of the signing CA
  - BOSH_CLIENT        - Username logged into BOSH
  - BOSH_CLIENT_SECRET - Password for BOSH_CLIENT

This does rely on the user logging in with a client-analog account,
which is most likely going to be admin.  This is a stopgap / hack so
that we can write hooks that don't rely on `~/.bosh/config` and aliased
environments.  Eventually, Genesis will pull these four variables out of
Exodus data for the BOSH environment that deployed us.

Fixes #277


dennisjbell:
We still use the bosh helper so that it knows the bosh executable, correct?

jhunt:
Yes.  This is strictly so that I don't also have to do `bosh -e $GENESIS_ENV_BOSH_TARGET` and pin Genesis to doing alias-based BOSH interaction moving forward.

PR #287, opened 2018-06-04
Remove old exodus data before submitting new after deploy

dennisjbell:
Now deletes existing exodus data after successful deployment to ensure outdated entries are not kept.


ci/release_notes.md
```diff
@@ -0,0 +1,4 @@
+# Bug Fix
```
jhunt: this should be "Bug Fixes"

PR #288, opened 2018-06-07
Fixed cloud-config handling for lookup --merged

dennisjbell:
Supports specifying a cloud config for `lookup --merged` using `--cloud-config` option, or by fetching cloud config from BOSH directory if not specified.


jhunt:
LGTM!

PR #289, opened 2018-06-12
_lookup now supports arrays and array lookup by key/value

dennisjbell:
This now works:

`genesis lookup --deployed my-env 'instance_groups[0].networks[name=public].static_ips'`

### DIFF
--- a/lib/Genesis/Env.pm
+++ b/lib/Genesis/Env.pm
@@ -224,7 +224,7 @@ sub _lookup {
 		if (/^(\d+)\]$/) {
 			return $default unless ref($what) eq "ARRAY" && scalar(@$what) > $1;
 			$what = $what->[$1];
-		} elsif (/^([^\[\=]*)=([^\]]*)]$/) {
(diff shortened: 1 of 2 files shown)


lib/Genesis/Env.pm
```diff
@@ -220,9 +220,25 @@ sub _lookup {
...
 	return $what if $key eq '';
 
-	for (split /\./, $key) {
-		return $default if !exists $what->{$_};
-		$what = $what->{$_};
+	for (split /[\[\.]/, $key) {
+		if (/^(\d+)\]$/) {
+			return $default unless ref($what) eq "ARRAY" && scalar(@$what) > $1;
+			$what = $what->[$1];
+		} elsif (/^([^\[\=]*)=([^\]]*)]$/) {
```
jhunt: is this more clearly articulated via non-greedy `.*` matches?  i.e.:

```
} elsif (/^(.*?)=(.*?)]$/) {
    # ...
}
```

PR #295, opened 2018-08-24
Fix `genesis embed` for Packed Genesii

### LINKED ISSUE #294: `genesis embed` copies wrong file
Currently, running `genesis embed` copies the existing Genesis runtime to `$pwd/.genesis/bin/genesis`. Because it is only the Genesis script itself, and none of the libraries, using `$pwd/.genesis/bin/genesis` will fail with 
```
Can't locate Genesis.pm in @INC (you may need to install the Genesis module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.26.1 /usr/local/share/perl/5.26.1 /usr/lib/x86_64-linux-gnu/perl5/5.26 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.26 /usr/share/perl/5.26 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at sb-changes/.genesis/bin/genesis line 23.
BEGIN failed--compilation aborted at sb-changes/.genesis/bin/genesis line 23.
```

A better method would be to copy the Genesis installation/inflation runtime (the file that can be downloaded via the GitHub releases). This way, when run, it installs the entire Geese runtime.

A temporary fix in the meantime would be to run `cp $(which genesis) ./.genesis/bin/genesis`, which embeds the Geese inflation script in the respective folder.

jhunt:
The GEESE stub sets GENESIS_CALLBACK_BIN so that hooks can call back to
the correct path.  We now also use that for embed, if it is set.

Fixes #294

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -628,7 +628,7 @@ sub {
 	check_prereqs;
 
 	# FIXME: update .genesis/config with new version info
-	Genesis::Top->new('.')->embed($0);
+	Genesis::Top->new('.')->embed($ENV{GENESIS_CALLBACK_BIN} || $0);
 });
 # }}}
 # genesis init - initialize a new Genesis repository {{{


Proplex:
LGTM. Tested on macOS, Ubuntu, and Concourse and everything is working as expected

PR #297, opened 2018-08-31
Added cloud_config_needs static_ip, and test

dennisjbell:
This adds the ability for kits to specify how many static_ips are needed, which is a better user experience than failing in spruce at manifest build time for each static IP that is missing.  Syntax is `cloud_config_needs static_ips <network_name> <quantity>`.  For example, the following directives from a mock cf-genesis-kit:

```
    cloud_config_needs static_ips $(lookup params.cf_internal_network cf-core) 15
    cloud_config_needs static_ips $(lookup params.cf_edge_network cf-edge) 25
```

might result in the following output:

```
[my-env] running cloud-config and environmental parameter checks...
Errors were encountered in your cloud-config:
 - network cf-core needs 15 static IP addresses, has 10
 - network cf-edge needs 25 static IP addresses, has 15
...
```

Like all cloud_config_needs call usage, this needs to be followed up with a block similar to the following to actually output the error messages and pass/fail:
```
  if check_cloud_config ; then
    describe "  cloud config [#G{OK}]"
  else
    describe "  cloud config [#R{FAILED}]"
    exit 1
  fi
```

Furthermore, the helper bash script that is used by the hooks is now checked with shellcheck, if present.  It has been added to the genesis Docker image in a previous direct-to-master commit.  It will fail on errors or warnings, but pass (and suppress) info and style level messages.  This is done to ensure any new potential errors are caught, but if it is deemed not an error, a directive can be added to the file to bypass that instance in the check (see https://github.com/starkandwayne/genesis/pull/297/files#diff-f57f19e6c402057bc4ea700b4515fdbaR417 for example)

### DIFF
--- a/lib/Genesis/Helpers.pm
+++ b/lib/Genesis/Helpers.pm
@@ -223,13 +223,18 @@ cloud_config_needs() {
 		__ips=$(spruce json "$GENESIS_CLOUD_CONFIG" | \
 			jq -r --arg network "$__network" '.networks[]| select(.name == $network) | .subnets[] | .static[]')
 
-		while read -r range ; do
-			read -r __f __x __l < <(echo "$range")
-			[[ -z "$__f" ]] && continue # blank line
-			if [[ -z "$__x" && -z "$__l" ]] ; then
-				(( __sum++ )) # Single ip
-			elif [[ "$__x" == '-' ]] ; then
-				(( __sum += $(__ip2dec "$__l") - $(__ip2dec "$__f") + 1 )) # Range
+		while read -r __range ; do
+			[[ -z "$__range" ]] && continue # blank line
+			if echo "$__range" | grep -q '^\s*[\.0-9]*\(\s*-\s*[0-9\.]*\)\{0,1\}\s*$' ; then
+				read -r __f __x __l < <(
+					echo "$__range" | \
+					sed 's/^[[:space:]]*\([^-[:space:]]*\)[[:space:]]*\(\(-\)[[:space:]]*\(.*\) \)\{0,1\}/\1 \3 \4/' | \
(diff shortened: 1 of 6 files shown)


lib/Genesis/Helpers.pm
```diff
@@ -196,10 +196,52 @@ export -f bosh_cpi
...
+	if [[ "${__type}" == "static_ips" ]] ; then
+		local __network=${1:?cloud_config_needs(static_ips) - must supply network name}; shift
+		local __count=${1:?cloud_config_needs(static_ips) - must supply static_ip count} ; shift
+
+		local __ips __sum=0 __f __x __l
+		__ips=$(spruce json "$GENESIS_CLOUD_CONFIG" | \
+			jq -r --arg network "$__network" '.networks[]| select(.name == $network) | .subnets[] | .static[]')
+
+		while read -r range ; do
+			read -r __f __x __l < <(echo "$range")
```
jhunt: IP ranges in BOSH cloud-config may not have whitespace around the delimiter.

For example:

```
subnets:
 - 10.0.0.1-10.0.0.15
```

is valid, and possible.

lib/Genesis/Helpers.pm
```diff
@@ -372,10 +414,10 @@ prompt_for() {
 		fi
 		if [[ $__type =~ ^multi- ]] ; then
 			eval "unset $__var; ${__var}=()"
-			local __i=0
+			# shellcheck disable=SC2034
```
jhunt: Can you put links, or some textual description of what is being suppressed?

dennisjbell:
Updated as per review notes, made modifications to issues found while testing on Linux boxes of trusty64 and xenial64

PR #298, opened 2018-09-13
Add the ability for Genesis to automatically extract AppRole from Exodus

Proplex:
This goes in hand with https://github.com/genesis-community/concourse-genesis-kit/pull/38 so that operators no longer have to extract AppRole credentials for pipeline setup.

### DIFF
--- a/lib/Genesis/CI/Legacy.pm
+++ b/lib/Genesis/CI/Legacy.pm
@@ -381,8 +381,6 @@ sub parse {
 	$P->{pipeline}{unredacted} = yaml_bool($P->{pipeline}{unredacted}, 0);
 
 	# some default values, if the user didn't supply any
-	$P->{pipeline}{vault}{role}   ||= "";
-	$P->{pipeline}{vault}{secret} ||= "";
 	$P->{pipeline}{vault}{verify} = yaml_bool($P->{pipeline}{vault}{verify}, 1);
 
 	$P->{pipeline}{task}{image}   ||= 'starkandwayne/concourse';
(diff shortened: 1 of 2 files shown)


lib/Genesis/CI/Legacy.pm
```diff
@@ -610,7 +610,9 @@ pipeline:
     repo:        (( param "Please specify the name of the Git repository" ))
     branch:      master
     private_key: (( param "Please generate an SSH Deployment Key and install it into Github (with write privileges)" ))
-
+  vault:
+    role:   (( vault "secret/exodus/ci/genesis_pipelines:approle_id" ))
+    secret: (( vault "secret/exodus/ci/genesis_pipelines:approle_secret" ))
```
jhunt: The outstanding item on the other related PR is to rename `genesis_pipelines` to `genesis-pipelines`; we'll need to remember to propagate the update here as well.

lib/Genesis/CI/Legacy.pm
```diff
@@ -1119,8 +1121,8 @@ EOF
             CACHE_DIR:            $alias-cache
             GIT_BRANCH:           (( grab pipeline.git.branch ))
             GIT_PRIVATE_KEY:      (( grab pipeline.git.private_key ))
-            VAULT_ROLE_ID:        $pipeline->{pipeline}{vault}{role}
-            VAULT_SECRET_ID:      $pipeline->{pipeline}{vault}{secret}
+            VAULT_ROLE_ID:        (( grab pipeline.vault.role ))
+            VAULT_SECRET_ID:      (( grab pipeline.vault.secret ))
```
jhunt: Since we're no longer using these, we should remove all the other references to handling them (in other parts of Legacy.pm), and warn if they are present.

Proplex: I've kept them here in case an operator wanted to override what we do. i.e. a security team doesn't like our policies or whatnot, and instead they'd prefer to manage their own AppRole/policies. By default, if `pipeline.vault.*` in `ci.yml` takes precedent if they're defined. Otherwise, fallback to our in-Vault, Concourse-kit-generated credentials.

Proplex:
Adjusted for feedback. I've removed references to `$P->{pipeline}{vault}{role}` and `$P->{pipeline}{vault}{secret}`.  I've also updated documentation that refers to the new addon command in the Concourse pipeline.

lib/Genesis/CI/Legacy.pm
jhunt: love it!

PR #307, opened 2018-10-03
Honour ca cert valid_for and names in kit.yml

dennisjbell:
Genesis hard-coded the ca cert validity to 1 year and ignored any
provided names, instead using a random name in the form of
`ca.n<random#>.<secret-name>`.  This solution applies the `valid_for`
property, and adds any provided names as Subject Alternative Name
entries, keeping the random name for the CN field to prevent conflict in
browsers due to reuse of CNs with different serial numbers.

### DIFF
--- a/lib/Genesis/Legacy.pm
+++ b/lib/Genesis/Legacy.pm
@@ -280,11 +280,19 @@ sub cert_commands {
 	my $missing_only = ($options{scope}||'') eq 'add';
 	for my $path (sort keys %$certs) {
 		my $rand = sprintf("%09d", rand(1000000000));
+		my $ttl = "1y";
+		my @names = ('--name', "ca.n$rand.$path");
+		if (defined($certs->{$path}{"ca"})) {
+			$ttl = $certs->{$path}{"ca"}{valid_for} || $ttl;
+			push(@names, '--name', $_)
+				for (@{ $certs->{$path}{"ca"}{names} || [] });
+		}
 		my @cmd = (
 			"x509",
 			"issue",
 			"secret/$options{prefix}/$path/ca",
-			"--name", "ca.n$rand.$path",
+			@names,
+			"--ttl" , $ttl,
 			"--ca");
 		push @cmd, "--no-clobber", "--quiet" if !$force_rotate; # All CA certs are considered kept
 		push @cmds, \@cmd;
--- a/t/secrets.t
+++ b/t/secrets.t
@@ -245,6 +245,29 @@ EOF
 	have_secret "$v/auto-generated-certs-a/ca:certificate";
 	my $ca = secret "$v/auto-generated-certs-a/ca:certificate";
 
+	sub get_cert_validity {
+		use Time::Piece;
+		my ($info) = @_;
+		my $pattern = "%b%n%d %H:%M:%S %Y";
+		my @i = $info =~ qr/Not Before:\s(.*\s+\d{4})\s+([^\n\r]*)\s+Not After\s+:\s(.*\s+\d{4})\s+([^\n\r]*)/m;
+		return undef unless $i[1] eq $i[3]; # ensure timezones are the same
+		return (Time::Piece->strptime($i[2], $pattern) - Time::Piece->strptime($i[0], $pattern));
+	}
+
+	# Check correct TTL
+	my $fixed_ca = qx(safe get $v/fixed/ca:certificate | openssl x509 -inform pem -text);
+	is get_cert_validity($fixed_ca), (5*365*24*3600), "CA cert has a 5 year validity period";
+
+	# Check CA alternative names and default TTL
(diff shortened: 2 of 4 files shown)


lib/Genesis/Legacy.pm
```diff
@@ -280,11 +280,19 @@ sub cert_commands {
 	my $missing_only = ($options{scope}||'') eq 'add';
 	for my $path (sort keys %$certs) {
 		my $rand = sprintf("%09d", rand(1000000000));
+		my $ttl = "1y";
+		my @names = ('--name', "ca.n$rand.$path");
+		if (defined($certs->{$path}{"ca"})) {
+			$ttl = $certs->{$path}{"ca"}{valid_for} || $ttl;
+			push(@names, '--name', $_)
+				for (@{ $certs->{$path}{"ca"}{names} || [] });
```
jhunt: This doesn't allow users to override the CN of the CA certificate, just the SANs.

t/secrets.t
```diff
@@ -245,6 +245,29 @@ EOF
 	have_secret "$v/auto-generated-certs-a/ca:certificate";
 	my $ca = secret "$v/auto-generated-certs-a/ca:certificate";
 
+	sub get_cert_validity {
+		use Time::Piece;
```
jhunt: Move this to the top of the test file (if we have to keep Time::Piece; see below)

t/secrets.t
```diff
@@ -245,6 +245,29 @@ EOF
...
+		my ($info) = @_;
+		my $pattern = "%b%n%d %H:%M:%S %Y";
+		my @i = $info =~ qr/Not Before:\s(.*\s+\d{4})\s+([^\n\r]*)\s+Not After\s+:\s(.*\s+\d{4})\s+([^\n\r]*)/m;
+		return undef unless $i[1] eq $i[3]; # ensure timezones are the same
+		return (Time::Piece->strptime($i[2], $pattern) - Time::Piece->strptime($i[0], $pattern));
+	}
+
+	# Check correct TTL
+	my $fixed_ca = qx(safe get $v/fixed/ca:certificate | openssl x509 -inform pem -text);
+	is get_cert_validity($fixed_ca), (5*365*24*3600), "CA cert has a 5 year validity period";
```
jhunt: Have you thought about using `safe x509 show` instead of relying on OpenSSL x509 utils?

```
$ safe x509 show secret/ssl/ca
secret/ssl/ca:
  cn=ca.ssl

  expires in 3350 days
  valid from Dec 8 2017 - Dec 6 2027 (~10 years)

  for the following purposes:
    (no special key usage constraints present)

  for the following names:
    - ca.ssl (DNS)
```

The `expires in` should match your TTL, without relying on Time::Piece.

t/secrets.t
```diff
@@ -245,6 +245,29 @@ EOF
...
+
+	# Check CA alternative names and default TTL
+	my $auto_b_ca = qx(safe get $v/auto-generated-certs-b/ca:certificate | openssl x509 -inform pem -text);
+	like $auto_b_ca, qr/Issuer: CN\s*=\s*ca\.n\d+\.auto-generated-certs-b/m, "CA cert is self-signed";
```
jhunt: To validate cert issuers, you can use the output of `safe x509 show` and check CNs.

t/secrets.t
```diff
@@ -245,6 +245,29 @@ EOF
...
+	like $auto_b_ca, qr/Subject: CN\s*=\s*ca\.n\d+\.auto-generated-certs-b/m, "CA cert is self-signed";
+	like $auto_b_ca, qr/Subject Alternative Name:\s+DNS:ca\.n\d+\.auto-generated-certs-b,\s+DNS:ca.asdf.com,\s+IP Address:127.1.2.3\s*$/sm,
+	               "CA has correct Subject Alternative Names";
```
jhunt: To check SANs, you can use `safe x509 validate <path> --for <name>` instead of OpenSSL

t/secrets.t
dennisjbell: I used the openssl because it is the base truth, not a derived fact, and all the existing tests use openssl.  I can probably switch to `safe x509 show` but not sure its worth it as this code already does what is needed.

dennisjbell: Time::Piece is core, so no CPAN module needed.

lib/Genesis/Legacy.pm
dennisjbell: This was due to the discussion that we don't want users supplying potentially breaking names, but we still want to support them being able to supply names in some meaningful fashion.

jhunt: I think we need to allow users to supply breaking names, if they want to.  We just need to randomize if missing, and then never set the CA name in a kit.

t/secrets.t
jhunt: We should change the existing tests as well.

PR #320, opened 2019-01-24
Bosh check timeout

dennisjbell:
Uses tcp_listening to check that the BOSH director is listening on the
specified host and port before blindly trying to connect to it.

### DIFF
--- a/t/05-bosh.t
+++ b/t/05-bosh.t
@@ -13,8 +13,6 @@ use Genesis;
 sub bosh_runs_as {
 	my ($expect, $output) = @_;
 	$output = $output ? "echo \"$output\"" : "";
-
-
 	fake_bosh(<<EOF);
 $output
 [[ "\$@" == "$expect" ]] && exit 0;
@@ -42,11 +40,13 @@ subtest '_bosh helper magic' => sub {
 
 subtest 'bosh ping' => sub {
 	local $ENV{GENESIS_BOSH_COMMAND};
+	my $director = fake_bosh_director('the-target');
 	bosh_runs_as("-e the-target env");
-	ok Genesis::BOSH->ping('the-target'), "bosh env should ping ok";
+	ok Genesis::BOSH->ping('the-target'), "bosh env on alias should ping ok";
 
 	bosh_runs_as("-e https://127.0.0.1:25555 env");
-	ok Genesis::BOSH->ping('https://127.0.0.1:25555'), "bosh env should ping ok";
+	ok Genesis::BOSH->ping('https://127.0.0.1:25555'), "bosh env on url should ping ok";
+	$director->stop();
 };
 
 subtest 'bosh create-env' => sub {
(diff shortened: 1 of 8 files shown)


jhunt:
This is more than just a tcp_listening change.  Let's discuss over a zoom.

dennisjbell:
@jhunt moved all the other changes into separate PRs; this one is now solely for the tcp_listening change.

bin/genesis
```diff
@@ -316,6 +316,8 @@ sub check_bosh_version {
 	my ($min) = @_;
 	my $best = "0.0.0";
 	foreach my $boshcmd (qw(bosh2 boshv2 bosh)) {
+		trace "HOME: $ENV{HOME}";
+		trace "BOSH Cmd %s found at '%s' (version output: %s)\n", $boshcmd, `which $boshcmd`, `$boshcmd -v 2>&1`;
```
jhunt: Isn't this trace a bit optimistic?  we might not have a bosh2 or a boshv2, but the trace output seems like it will imply that there is one.

lib/Genesis.pm
```diff
@@ -710,6 +710,18 @@ Debug messages are all prefix with the string "TRACE> ".
...
+Dumps one or more named values to standard error if C<$GENESIS_TRACE> or
+C<$GENESIS_TRACE> environment variables have been set to "truthy".  Optional
+scope level of -1 or less will report the corresponding stack level as the
+source of the output, defaults to the calling scope.
+
+=head2 dump_stack()
+
+Dumps the current stack to standard error if C<$GENESIS_TRACE> or
+C<$GENESIS_TRACE> environment variables have been set to "truthy".
+
```
jhunt: Are these from a previous PR?

t/05-bosh.t
```diff
@@ -10,11 +10,10 @@ use Test::Output;
 use_ok "Genesis::BOSH";
 use Genesis;
 
+
```
jhunt: superfluous whitespace?

t/30-env.t
```diff
@@ -624,16 +628,21 @@ EOF
 
 	{
 		$env = $top->load_env('standalone');
-		local $ENV{GENESIS_BOSH_ENVIRONMENT} = "https://127.0.0.86:25555";
+		local $ENV{GENESIS_BOSH_ENVIRONMENT} = "https://127.0.0.1:26666";
 		$env = $top->load_env('standalone'); # reload otherwise its cached by the previous call
-		is $env->bosh_target, "https://127.0.0.86:25555", "the \$GENESIS_BOSH_ENVIRONMENT overrides all";
+		is $env->bosh_target, "https://127.0.0.1:26666", "the \$GENESIS_BOSH_ENVIRONMENT overrides all";
```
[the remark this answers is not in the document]
dennisjbell: Yes, that is the address of the fake_bosh_director, so it needs to match that.

bin/genesis
dennisjbell: leftover debugging -- has been removed

lib/Genesis.pm
dennisjbell: Yes, moved into master

t/05-bosh.t
dennisjbell: yup, removed

PR #322, opened 2019-02-15
Add the --target if a GENESIS_TARGET_VAULT is set

krutten:
When exodus is called with the GENESIS_TARGET_VAULT the command line was missing the `--target` option which confuses safe

```
++ local __env __key
++ __env=hplab/cf
++ __key=autoscaler_enabled
++ [[ -n '' ]]
++ safe https://10.250.4.48 exists secret/exodus/hplab/cf:autoscaler_enabled
Valid commands are:

    ask         Create or update an insensitive configuration value
    auth        Authenticate to the current target
    copy        Copy a secret from one path to another
    curl        Issue arbitrary HTTP requests to the current Vault (for diagnostics)
    delete      Remove one or more path from the Vault
    dhparam     Generate Diffie-Helman key exchange parameters
    env         Print the environment variables for the current target
    exists      Check to see if a secret exists in the Vault
```

We now add `--target` if GENESIS_TARGET_VAULT is set

### DIFF
--- a/lib/Genesis/Helpers.pm
+++ b/lib/Genesis/Helpers.pm
@@ -111,8 +111,8 @@ exodus() {
     __key=$2
     __env=$1
   fi
-  if safe "$GENESIS_TARGET_VAULT" exists "secret/exodus/${__env}:${__key}"; then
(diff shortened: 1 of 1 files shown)


jhunt:
@dennisjbell did you have a chance to look at this?

PR #323, opened 2019-03-04
Update readme with CentOS/Amazon Linux dependencies

daviddob:


dennisjbell:
This doesn't make sense -- those are all core libraries, included within Perl (for example Data::Dumper: https://perldoc.perl.org/Data/Dumper.html )

daviddob:
While I agree, and they should be included, on Amazon Linux and CentOS they are not included with the perl packages. (Confirmed on amzn-ami-hvm-2018.03.0.20181129-x86_64-gp2 (ami-47d4b326) today)

dennisjbell:
Agreed, as this is just a readme.

PR #328, opened 2019-04-10
Support env file min_version of genesis

dennisjbell:
Genesis environment files can now specify the minimimum version of
Genesis that is required to read it.  This is automatically populated
when kits use the new `genesis_config_block` helper and specify a
minimum verison required by the kit, or setting `$GENESIS_MIN_VERSION`
prior to calling that helper..  It can also be expicitly stated by
specifying `genesis.min_version` or `params.genesis_version_min` in the
environment file.

Furthermore, the deprecation warning for not using `genesis` block for
the values in `params` that were moved to `genesis` have been delayed to
a future version, so that detection of kit and env depencency checks can
be in the wild for some time.

### DIFF
--- a/ci/release_notes.md
+++ b/ci/release_notes.md
@@ -133,7 +133,7 @@ domain names that can be used as the host.
   so Genesis will auto-populate the `params.env` for them, ensuring that any
   reliance on this does not break existing kits.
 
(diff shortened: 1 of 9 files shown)


dennisjbell:
No, those changes were never released.

dennisjbell:
Rebased for merging

PR #330, opened 2019-04-15
Add edit_environment helper with man support

### LINKED ISSUE #324: Add edit_environment helper with built-in manual
There is a growing trend for kits to allow the generated environment file to be edited prior to secrets generation or other actions.  I would like to capture this in a helper that opens a vim session split with a read-only copy of the manual on the right side, so that a) users know the options that they can change, b) users don't have to open the web or another shell to see the manual, and c) users learn that you can actually get the manual from genesis.

dennisjbell:
vim and emacs support, other editors don't get the manual

Fixes #324

### DIFF
--- a/lib/Genesis/Helpers.pm
+++ b/lib/Genesis/Helpers.pm
@@ -514,3 +514,25 @@ EOF
 	echo ""
 }
 export -f genesis_config_block
+
+edit_environment() {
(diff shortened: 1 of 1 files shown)


dennisjbell:
This was accidentally merged into master -- killing this PR and reopening

PR #335, opened 2019-04-30
Update init help with --vault

### LINKED ISSUE #333: Genesis init help does doesn't include vault options
In the new secrets provider code, Genesis prompts on `init`

```
Setting up Genesis *-deployments/ directory...

[installing genesis runtime]

Select Vault:

  1) da-vault   (noverify) https://10.200.130.4 (default)

Select choice > 
```

which broke automation that doesn't expect this.  The help for `init` doesn't show a method to pre-specify a vault

```
$ genesis init --help
genesis v2.6.13 (05885f92f6) build 20190416.033706
USAGE: genesis init [-k KIT/VERSION] [-d directory] [name]

OPTIONS
  -h, --help        Show this help screen.
  -D, --debug       Enable debugging, printing helpful message about what
                    Genesis is doing, to standard error.
  -T, --trace       Even more debugging, including debugging inside called
                    tools (like spruce and bosh).
  -C, --cwd         Effective working directory.  Defaults to '.'
  -e, --environment Which BOSH environment (aka director) to use.  If not
                    specified, it will use the value in params.bosh or
                    params.env in that order.  Can also be provided using either
                    $GENESIS_BOSH_ENVIRONMENT or $BOSH_ENVIRONMENT env variables.

  -k, --kit           Name (and optionally, version) of the Genesis Kit to base
                      these deployments on.  I.e.: shield/6.3.0.  If you do not
                      specify a kit, a dev directory will be created for you to
                      develop a local kit into.
  -L, --link-dev-kit  Instead of using a kit or initializing an empty dev
                      directory, this will link the specified directory to the
                      dev directory.
  -d, --directory     By default, the directory in which the Genesis deployment
                      will be created in will be named ./<name>-deployments.
                      Use this option to change it to something else.

  name                If the name argument is not specified, it will default to
                      the same name as the kit.  You must specify either name
                      or kit.
$
```

thomasmitchell:
halp.

<.< ... Fixes #333

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -665,6 +665,9 @@ $GLOBAL_USAGE
   -L, --link-dev-kit  Instead of using a kit or initializing an empty dev
                       directory, this will link the specified directory to the
                       dev directory.
+      --vault         Name of a `safe' target (a Vault) to store newly generated
+                      credentials in. If not specified, it will be interactively 
+                      prompted for.
   -d, --directory     By default, the directory in which the Genesis deployment
                       will be created in will be named ./<name>-deployments.
                       Use this option to change it to something else.
--- a/bin/genesis
+++ b/bin/genesis
@@ -665,7 +665,7 @@ $GLOBAL_USAGE
   -L, --link-dev-kit  Instead of using a kit or initializing an empty dev
                       directory, this will link the specified directory to the
                       dev directory.
-      --vault         Name of a `safe' target (a Vault) to store newly generated
+      --vault         Name of a 'safe' target (a Vault) to store newly generated
                       credentials in. If not specified, it will be interactively 
                       prompted for.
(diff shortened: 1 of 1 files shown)


bin/genesis
```diff
@@ -665,6 +665,9 @@ $GLOBAL_USAGE
   -L, --link-dev-kit  Instead of using a kit or initializing an empty dev
                       directory, this will link the specified directory to the
                       dev directory.
+      --vault         Name of a 'safe' target (a Vault) to store newly generated
```
jhunt: Can we write this with a leading backquote?  as in:

```
Name of a `safe' target ...
```

dennisjbell: Really....?

thomasmitchell: It's called a "Grave Accent``

PR #345, opened 2019-06-18
Remove reliance on logsearch kit existing

dennisjbell:
The tests for downloading kits expected logsearch kit to exist, but it doesn't anymore.


thomasmitchell:
This is perfect.

PR #347, opened 2019-06-19
Fix cloud config check typo

thomasmitchell:


dennisjbell:
You're misunderstanding the purpose -- it is a checklist of what is expected, but the red x means it was not met.  All checks are stated in the positive.

jhunt:
Sometimes the X doesn't render (especially under tmate / tmux); so having the change of text is helpful.

PR #337, opened 2019-05-14
drop v1 support

dennisjbell:
Genesis v1 has been dropped for many years now, with no further updates.  This change, to be released at v2.7.0, will drop the embedded genesis v1, in an effort to shrink the Genesis package size.

PR #351, opened 2019-07-19
Drop Support for v1 Genesis

dennisjbell:
(This is a repeat of pull request #337, which was backed out of the master branch due to the need to continue minor updates to the 2.6.x line.  It will need to be re-merged once we are ready to proceed with v2.7.0)

Genesis v1 has been dropped for many years now, with no further updates. This change, to be released at v2.7.0, will drop the embedded genesis v1, in an effort to shrink the Genesis package size.


dennisjbell:
This was unmerged to get something out, then remerged as #352

PR #376, opened 2020-03-25
Fix error handling with run and load_yaml_file

dennisjbell:
There is a poorly understood paradigm where if you set stderr to 0 on
run calls, it would just keep stderr on stderr instead of the default of
merging it with stdout.  This meant stderr output from the run calls
would just appear on the user's terminal, which doesn't give Genesis
much control of the occurrence or handling.

The first attempt to resolve this changed the behaviour of `{stderr => 0}`
to just drop stderr into /dev/null (3b0473d), which solved the issue of the
output, but it didn't give any better control, and the user didn't get
to see what the issue was.

Instead of just sending stderr to /dev/null or the terminal, this change
captures stderr to a file, then reads that file in at the conclusion of
the run being completed.  Depending on the other options, it will log
the stderr to the trace log, and either print it out if errors were
marked as fatal, or return it as a third element after $out and $rc.

There are still cases where this output is hidden:

* passfail is enabled:
    This seems acceptable, even preferable, as stdout is also dropped and the
    intent of that option is to only care about the return code.

* onfailure is enabled, and  STDERR is written to, but the command exited
  with rc of 0:
    This could be problematic in rare circumstances, none of which come
    to mind

* calling run in a scalar or void context:
    This is perlism for assigning the result of the run call to a scalar
    variable (`my $out = run(...)`) or not doing anything with the
    output (`run(...)`).  In this case, the controlling code doesn't get
    it, and it didn't get put on the terminal, so it is lost.

There are only so many cases of `{stderr => 0}` in the code, with curl
and load_yaml_file being the primary usage.  Beyond that most usage of
run merges the stderr into stdout (default) or more esoteric usage.

Also:
- Fix encoding on error output for when UTF-8 characters are used.
- Improve stack dump scoping on bug report

### DIFF
--- a/lib/Genesis.pm
+++ b/lib/Genesis.pm
@@ -256,6 +256,7 @@ sub _log {
 
 sub error {
 	my @err = @_;
+	binmode(STDOUT, "encoding(UTF-8)");
 	unshift @err, "%s" if $#err == 0;
 	print STDERR csprintf(@err) . "$/";
 }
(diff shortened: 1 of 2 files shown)


krutten:
In discussing the change with Dennis, a better solution for `stderr => 0` may be to collect the STDERR and pass it to the calling function on failure.

PR #387, opened 2020-03-28
Add credhub support

dennisjbell:
Adds a credhub bash helper function for hooks, and the ability to specify that an environment uses credhub (so don't do the vault validations).

The credhub helper requires an environment that provides the connection information via exodus data.  The latest BOSH genesis kit contains these values.  By default, the bosh environment name will be used, but can be overrode by using `genesis.credhub-exodus-env`

### DIFF
--- a/lib/Genesis/Kit.pm
+++ b/lib/Genesis/Kit.pm
@@ -257,7 +257,7 @@ sub metadata {
 # uses_credhub - does this kit use credhub instead of vault {{{
 sub uses_credhub {
 	my ($self) = @_;
-	return defined($self->metadata->{secret_store}) && $self->metadata->{uses_credhub} eq "credhub";
+	return defined($self->metadata->{secret_store}) && $self->metadata->{secret_store} eq "credhub";
 }
 
 # }}}
--- a/lib/Genesis/Kit.pm
+++ b/lib/Genesis/Kit.pm
@@ -100,6 +100,17 @@ sub run_hook {
 		}
 
 		unless (grep { $_ eq $hook } qw/new prereqs/) {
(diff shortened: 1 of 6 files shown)


lib/Genesis/Kit.pm
```diff
@@ -242,6 +253,13 @@ sub metadata {
 	return $self->{__metadata};
 }
 
+# }}}
+# uses_credhub - does this kit use credhub instead of vault {{{
+sub uses_credhub {
+	my ($self) = @_;
+	return defined($self->metadata->{secret_store}) && $self->metadata->{uses_credhub} eq "credhub";
```
krutten: second value should also be "uses_credhub"

lib/Genesis/Kit/Compiler.pm
```diff
@@ -68,7 +68,18 @@ sub validate {
 		}
 
 		# check for errant top-level keys - params, subkits and features have been discontinued.
-		my @valid_keys = qw/name version description code docs author authors genesis_version_min certificates credentials/;
+		my @valid_keys = qw/name version description code docs author authors genesis_version_min secrets_store/;
+		if ($meta->{secrets_store}) {
+			if ($meta->{secrets_store} eq "credhub") {
+				#no-op: valid, no further keys
+			} elsif ($meta->{secrets_store} eq "credhub") {
```
krutten: credhub should vault

lib/Genesis/Kit.pm
dennisjbell: fixed

lib/Genesis/Kit/Compiler.pm
dennisjbell: fixed

PR #392, opened 2020-04-02
V2.7.0 rc4

dennisjbell:
Massive update to how Genesis handles kit secrets.  This is done using
the new functionality provided by Genesis::Vault (see below) in better handling kit
secrets generation and validation.

Updates include:

* New parameters for targeting secrets under the `genesis:`
  key in the environment YAML file:
  * `secrets_mount` specifies the base path in the vault under which all
    kit secrets are generated and accessed.  The existing
    `secrets_path`, also previously known as `vault_path` or `vault`,
    will be appended on the end of this.  By default, it is still
    `/secret/`
  * `exodus_mount` specifies the base path where the exodus data will be
    found.  By default it is `$secrets_mount/exodus/`
  * `ci_mount` specifies where any CI (Concourse for now) secrets are
    written to, and by default is `$secrets_mount/ci`
  * `root_ca_path` allows you to specify a path to a root CA certificate
    that will be used to sign all otherwise-unsigned certificates in the
    kit.  This allows you to distribute this CA as a source of
    authority.
  * Only kits that support v2.7.0-rc1 or higher will be allowed to use
    these additional parameters

* `genesis new` supports all of these as options, with double-dash
  prefix and convert underscores to dashes.  i.e.: to set
  `genesis.secrets_mount` during `genesis init`, specify
  `--secrets-mount`

* Addition of `genesis remove-secrets` to remove all (or just some) of
  the secrets specified by the kit.  This will not remove secrets that
  are stored during the environment creation unless you specify the
  --all flag, which can protect against undesired losses of ungenerated
  secrets.

* `genesis check-secrets` now supports `--validate` to check  the
  validity of your secrets instead of just presence.

* `genesis rotate-secrets` now supports `--renew` so you can renew the
  expiry date of your X509 certificates without invalidating their
  signing chains.

* Fine-tune selection on all *-secrets commands with -F|--filter option,
  which allows you to specify a regular expression to target just the
  secrets you want to act upon.  Further selection for rotate and
  remove is the -X|--failed option that allows you to just rotate or
  remove those secrets that are invalid.

* More feedback, less verbosity (or more if you want it).  When
  processing `*-secrets` commands, the minimum screen real estate will
  be used, reusing the same line as it progresses through each secret,
  reporting only issues.  However, if you want the full details, you can
  specify the -v|--verbose flag and get all the information points
  available.

------

### Genesis::Vault now contains routines to handle better secrets generation
and validation.  These improvements include:

* Analizes kit secrets (contents of kit.yml) -- This was refactored from
  Genesis::Legacy
  * Supports additional features such as `is_ca` and `signed_by` so you
    can specify arbitrarily named x509 secrets as CA certs and specify
    which CA signs another certificate, allowing multi-level chains.
  * Supports a common root CA path so that all unspecified CA certs will
    be signed by that common CA
  * Determines the order to create the certificates in to ensure CAs
    will always be present for signing - detects cycles
  * Adds support for `usage` array to specify certificate usage

* Supports renewing (updating TTL) and removal of secrets in addition to
  adding or recreating.

* Can validate (instead of just detect presence) of all secrets against
  the specifications in kit.yml, and provides a clear report of what
  errors were found. Useful when kits are upgraded and secrets change
  their specification.

* Improved output support (in conjunction with a provided callback
  function) that provides users better progress report as it processes
  kit secrets requests.

* Much faster processing of secrets presence and validation.

* Can fetch all secrets for a given environment, ensures they are not
  written to disk or logged.

* Can validate kit.yml prior to compiling the kit to catch errors before
  end users are faced with a kit that cannot generate the correct
  secrets.

Other adjunct improvements are:

* Better tracking of current Vault object by way of an ID (to prevent
  same value target, but loss of internal state)

* Better reconnection handling that ensures vaults are current.

-----

### Refactor `genesis compile-kit` significantly:

* Moved porcelain check of git repo inside validate, so that other
  checks would also happen
* Validate on viability of the kit secrets to ensure kit author didn't
  create a kit that will not create viable secrets
* Validates any parameter dereferencing

-----

### Rename --environment to --bosh-env:

The term 'environment' is overloaded in Genesis, so to reduce confusion,
the `--environment` option that was used to specify the bosh environment
to use has been renamed `--bosh-env`.  Its shorthand is still `-e` and
`--environment` is still honoured, but deprecated.

-----

Various other minor improvements and bug fixes


dennisjbell:
v2.7.0-rc5 supersedes this PR

PR #403, opened 2020-04-18
enhance test env setup instructions for linux

VasylTretiakov:
added the list of supported Linux distributions based on the contents of jumpbox script

### DIFF
--- a/t/README.md
+++ b/t/README.md
@@ -16,9 +16,8 @@ On OSX:
   * `brew install jq`
   * install install from https://www.vaultproject.io/downloads.html
 
-On Linux:
-  * The most straight-forward method of installing under linux is to install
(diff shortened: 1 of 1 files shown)


jhunt:
Which Linux are you trying to install on, @VasylTretiakov?

I'd rather update the instructions, rather than clarify the exclusion.

VasylTretiakov:
In fact, my point is that the current instructions do not specify which Linux distributions are supported by the jumpbox script. So my suggestion is to list those distributions in the instructions instead of letting someone potentially fail while trying to use it on an unsupported Linux distribution.
Quick look at the script suggested it covers three package managers used by the distributions in the suggested edit.
I do not think I am suggesting anything that would emphasize an exclusion.

PR #406, opened 2020-04-19
add kit commands to kit dev paragraph

VasylTretiakov:
Added `create-kit`, `compile-kit` and `decompile-kit` commands to kit development instructions.

### DIFF
--- a/docs/AUTHORING-KITS.md
+++ b/docs/AUTHORING-KITS.md
@@ -120,6 +120,9 @@ doesn't use `dev/`.  These are the ground rules:
 
 In dev-mode, the `kit.version` parameter is ignored.
 
+Genesis comes with `compile-kit` and `decompile-kit` commands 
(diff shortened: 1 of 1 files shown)


jhunt:
I'm struggling with this PR.  While it is factually accurate (Genesis does provide those three commands) it doesn't seem like simply listing them off does much good.  The proposed 2-line change doesn't really add much else to the state of developing kits.

Is it safe to assume that you came to the Authoring Kits document looking for something about compile-kit / decompile-kit and/or create-kit, and didn't find what you needed, @VasylTretiakov ?

VasylTretiakov:
Indeed, I was looking for a way to make Dev kit mode work, but failed miserably.
While reviewing Genesis help I noticed the discussed commands with at least some usage guidance.
Of course, I would prefer to have more insightful guidelines on using kit development commands provided instead.

PR #419, opened 2020-07-22
fix typo for description

7hunderbird:
quick typo fix


dennisjbell:
The misspelling of description is intentional in the test; its to ensure that it gets caught by the error handler.

PR #428, opened 2021-02-12
Add support for Credhub-based secret management

dennisjbell:
Genesis started as a vault-based system, but since then, Credhub has been utilized to store and generate secrets for some kits, most notably the cf-genesis-kit.  It is not a 1:1 replacement for vault, with significant differences in storage and behaviour. However, we want to be able to continue using genesis primitives to manage secrets in the same way.

To this end, the `lib/Genesis/Credhub.pm` library will attempt to bridge the differences to form a common vocabulary.  This will include, but not limited to:

1) using `variables` from the rendered manifest instead of kit.yml to define the secrets behaviour
2) check script will only error on missing credhub values if they are not specified by `variables` or `bosh_variables` sections
3) rotate secrets will use regenerate, but will have to handle changes to the manifest (ie if I change the password length or certificate properties, how does that get sent to credhub prior to regeneration without deploying the manifest.

`lib/Genesis/Vault.pm` defines the general behaviour, and can be used as a guide.  As it will mainly be done as running credhub cli commands, you'll want to be familiar with `run`, `read_json_from`, and `lines` subroutines in `lib/Genesis.pm`

### DIFF
--- a/lib/Genesis/Credhub.pm
+++ b/lib/Genesis/Credhub.pm
@@ -1,0 +1,1559 @@
+package Genesis::Credhub;
+use strict;
+use warnings;
+
+use Genesis;
+use Genesis::UI;
+use JSON::PP qw/decode_json/;
+use UUID::Tiny ();
+
+### Class Variables {{{
+my $current_credhub;
+# }}}
+
+### Class Methods {{{
+
+# new - raw instantiation of a credhub object {{{
+sub new {
+  my ($class, $name, $url, $username, $password, $ca_cert) = @_;
+  return bless({
+    name     => $name,
+    url      => $url,
+    username => $username,
+    password => $password,
+    ca_cert  => $ca_cert,
+  }, $class)
+}
+
+# }}}
+# attach - builder for vault based on loaded environment {{{
+sub attach {
+  my ($class, $name, $url, $username, $password, $ca_cert) = @_;
+  my $credhub = $class->new($url, $username, $password, $ca_cert);
+  $credhub->connect_and_validate;
+	$current_credhub = $credhub;
+	return $credhub;
+}
+
+# }}}
+# rebind - builder for rebinding to a previously selected vault (for callbacks) {{{
+sub rebind {
+}
+
+# }}}
+
+# current - return the last vault returned by attach, target, or rebind {{{
+sub current {
+	return $current_credhub;
+}
+
+# }}}
+# clear_all - clear all cached data {{{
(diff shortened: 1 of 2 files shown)


rkoster:
Collecting some research notes on for implementation direction:

Cf-deployment specifies `update_mode` for at least some variables: https://github.com/cloudfoundry/cf-deployment/blob/main/cf-deployment.yml#L2070

The bosh director has support for forwarding this information to credhub: https://www.pivotaltracker.com/n/projects/956238/stories/164818604
And is used here: https://github.com/cloudfoundry/bosh/blob/e0ac496c97eff51c3c7071928f2b96e4b347847c/src/bosh-director/lib/bosh/director/config_server/client.rb#L89-L116

Credhub api docs are currently down but below a screenshot from a local copy:
image

Credhub cli go client supports converge: https://github.com/cloudfoundry-incubator/credhub-cli/commit/37f2848fb10ab1b55cd568424e407b05654f4925 

Credhub cli `generate` command has `--no-overwrite` flag, no --mode or similar to set converge, `regenerate` has nothing like that. But if we already have all generation options from the manifest we should be able to use `generate`.

https://bosh.io/docs/manifest-v2/#features

Warning
The converge_variables feature can result in unintended side effects. It will regenerate all variables associated with a deployment, even those that are specified in a runtime config. For this reason we strongly recommend that you use the update_mode configuration on the variable that you wish to converge.

From the above it seems like the following requirements could be handled by BOSH using `converge_variables` and `update_mode`:

- [x] 1. using variables from the rendered manifest instead of kit.yml to define the secrets behaviour _by setting `update_mode` to no-overwrite for `"fixed"` secrets and CA's and `converge` on all others_
- [x] 2. check script will only error on missing credhub values if they are not specified by variables or bosh_variables sections
- [ ] 3.1 rotate secrets will use regenerate
- [x] 3.2 but will have to handle changes to the manifest (ie if I change the password length or certificate properties, how does that get sent to credhub prior to regeneration without deploying the manifest.) _This is exactly what update_mode converge was designed for_

rkoster:
3.1 could be implemented using the `bosh variables` cli command in combination with `bosh manifest` (to get the variable update_mode, so we can skip no-overwrite and isCa) and `bosh config` (to get runtime-config variable definitions).

The above should work just fine for the simple cases, but does not fix the 3 phase deploy required for rotating ca's.
Or the cc db encryption key, which requires an errand. Or the application ca used by diego to create certs on the fly.

PR #430, opened 2021-02-23
check for credhub version fixes #425

### LINKED ISSUE #425: Add credhub cli version to dependencies of genesis
The same way bosh, safe, jq, spruce and others are required dependencies, with minimum versions, the credhub cli is now at the point where it is no longer optional.

ramonskie:
fixes #425
i could not find  any tests for checking min versions


dennisjbell:
Add release notes to ci/release_notes.md

PR #431, opened 2021-02-24
Allow renew secrets to update metadata

dennisjbell:
`genesis rotate-secrets --renew` can now update the subject, subject
alternative names, and key usage in addition to resetting the
time-to-live value.


krutten (approved):
Reviewed with Dennis.  LGTM

Issue #432, opened 2021-03-03
suppress ci notify warning when non-cached files in changes dir are same as cache dir


thomasmitchell:
As it currently stands, we "fail" the notify job in CI when there is a file present in cache that is not in changes, or vice versa. This is to prevent situations where Concourse has updated one side and not the other, and the user has shifted contents from a more global file to a less-global file, or vice versa. If that were to happen, we could either find ourselves missing or duplicating configuration, respectively.

However, for a lot of cases when a shared configuration file is added, it is not to shift configuration out of a leaf - it is simply to add a new configuration option that should be shared by multiple environments. In this case, `changes` in the pipeline never updates, and continues to fail the notify job until the `changes` file is eventually updated for a different reason. Red Concourse jobs make people sad.

We can suppress this warning in cases where we can detect that configuration has actually not been shifted to a new location and, instead a file has simply been added or removed. We can do this detection by comparing non-cached files in the changes directory with those of the cached directory. If they differ, we can conclude that this is the result of desyncing in Concourse resource polling, and we should warn as we always have. If they are the same, then no desync can have occurred, and we should not warn (both because the warning is a false-positive, and also because it will warn indefinitely).

thomasmitchell:
After discussion, when files are removed  this should still warn, because genesis will try to use those files from the changes dir (when it shouldn't).

PR #439, opened 2021-03-05
Suppress extraneous extra files warning in CI

thomasmitchell:
Previously, notify jobs in genesis ci had a large window for false
positives for the warning when an environment file was added to the
repo. Reined that in a little bit, by not warning if non-cached files
are identical.


thomasmitchell:
This addresses #432

PR #440, opened 2021-03-06
Fix empty GIT_GENESIS_ROOT env var

dennisjbell:
Suppresses error message for undefined value error on GIT_GENESIS_ROOT
environment variable when checking if its set.

[//]: <> "category:Bug Fixes"


mrferris:
Closing because this commit was already merged in and github doesn't seem to realize it.

PR #450, opened 2021-05-27
Improve debugging logs

dennisjbell:
Adds the -S|--show-stack debug option, and gives better context of traces
and errors.

Behaviour Changes:

- internal `trace` output will show the location in the code were it was
  called. If --show-stack option specified, it will show the full stack
  instead of just the location.

- internal `dump_vars` output will now show the full stack instead of
  just the calling location if --show-stack specified (along with
  --trace or --debug)

- internal `bail` will show calling location if --trace or --debug is
  specified, or the full stack if --show-stack is specified (with or
  without --debug or --trace).  This allows you to see where genesis
  terminates without wading through all the trace output.

- internal `bug` will always full stack where the bug was encountered.
  This allows the user to paste the stack into a defect report against
  the genesis project.

### DIFF
--- a/lib/Genesis.pm
+++ b/lib/Genesis.pm
@@ -167,6 +167,7 @@ sub _glyphize {
 		' ' => '  ',
 		'>' => "⮀",
 		'!' => "\x{26A0} ",
+		'^-' => "\x{2B11} ",
 	);
 
 	$glyph = $glyphs{$glyph} if !envset('GENESIS_NO_UTF8') && defined($glyphs{$glyph});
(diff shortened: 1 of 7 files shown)


lib/Genesis.pm
```diff
@@ -268,17 +260,47 @@ sub _log {
...
+	}
+	chomp $out;
+	return $out;
+}
+
+sub _get_stack {
+	my ($scope) = @_;
+
+	my ($file,$line,$sub,@stack,@info);
+  while (@info = caller($scope++)) {
```
mrferris, line 296: indentation got a little off

Issue #449, opened 2021-05-24
Genesis concourse deployment of proto bosh mismanages state file


mrferris:
The bosh deploy of protobosh appears to be incorrectly using the state file from changes instead of the cache only in the specific case of a protobosh deploy run by the CLI's ci-deploy-changes command in a CI pipeline.

PR #460, opened 2021-06-14
Merging fix for issue #449, initial proto bosh deploy fails

mrferris:
Potential fix for #449 that is verified to leave non-initial-protobosh behavior unaffected. Also includes improvements to logging and shellcheck use during testing.


mrferris:
fixes #449

PR #476, opened 2022-06-03
Use /usr/bin/env for shebangs

patryk-kozak:
Modifies all of the source files that refers to either `bash` or `perl` using `/usr/bin` or `/bin` in the shebangs.
Shebangs like that make assumption that program will be found under a specified path, whereas in proposed change the `env` is used to actually localize the path of the given program. 

It will make CLI tool more portable as it should be runnable from almost any OS/Unix distribution since different *nix based systems.

Tested by running locally `perl pack` which generates the dev-based package of `genesis` tool.

https://stackoverflow.com/a/10383546
https://stackoverflow.com/a/16365367
https://stackoverflow.com/a/2792076

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1,4 +1,4 @@
-#!/usr/bin/perl
+#!/usr/bin/env perl
 # vim:ft=perl:noet:ts=2:sts=2:sw=2:fdm=marker:fdl=0
 use strict;
 use warnings;
@@ -327,7 +327,7 @@ EOF
 	if ($ENV{GIT_USERNAME}) {
 		mkdir_or_fail( "$env_dir/home", 0700);
 		mkfile_or_fail("$env_dir/home/credential-helper.sh", 0755, <<'EOF');
-#!/bin/bash
+#!/usr/bin/env bash
 echo username=$GIT_USERNAME
 echo password=$GIT_PASSWORD
 EOF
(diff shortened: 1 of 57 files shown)


norman-abramovitz:
Just some general comments about using the `env` command.

The `env` command uses the PATH variable by default to find the application it wants to run.  
1. There is an assumption that the `env` command is installed.  It is most likely is but there is no guarantee.
2. Since the PATH variable is controlled by the user, the user can choose which version of the application will be run.
     a.  This means different users could be running different versions of the application.
     b.  There could be incompatibilities between the application versions.
     
Using the `env` command is great for development work, but not so great for support teams when issues pop up.

patryk-kozak:
> 1. There is an assumption that the `env` command is installed.  It is most likely is but there is no guarantee.
I think if we actually think in that categories, there is no guarantee that bash or perl will be there as well. I would give more credit that `/usr/bin/env` will be on OS, then for ex. `bash` shell (for ex. Ubuntu default shell is now `dash`, not `bash`, for macOS since Catalina I think it is `zsh`, not `bash`, there are symlinks but they are not compatible when using their features).

> 2. Since the PATH variable is controlled by the user, the user can choose which version of the application will be run.
>    a.  This means different users could be running different versions of the application.
>    b.  There could be incompatibilities between the application versions.
That's very true, but also I don't think that's a bad thing unless we specify which specific interpreters, in which version, and where needs to be present on the user machine.
From what I read on README.md it just assumes user has Perl - it doesn't specify the required version or the required path it should be located under.

> Using the `env` command is great for development work
Hmmm, as it may look like that at first, I don't think I can fully agree on that. 
If the goal of genesis cli is to be portable and runnable under any OS, or at least majority of OS'es - it should not be left for users choice to run it through their local machine tooling, it should be distributed as self-contained binary with no dependencies, or with the env. Right now it is not, as there is no way to limit what version of Perl or other dependencies user have or have not, it also assumes the dependencies will be provided by the user.
Also: using `env` as some middle ground to solve shebang's problem is the standard practice for a many great open-source tools, including but not limiting to:
- https://github.com/kubernetes/kubernetes
- https://github.com/cloudfoundry/cli
- https://github.com/cloudfoundry/bosh
- https://github.com/golang/go
It also shows when searching the entire open-source GitHub code for `bash` use-cases 71k results vs 2M results in favor of using `/usr/bin/env bash`.

I think it may be OK to retract a change for the *test based source code, if the team feels it is better to control it by pointing to specific interpreters (for the sake of CI/CD/testing flows there are), it seems to be actually good thing if the infrastructure is tighly controlled and specialized.

>  but not so great for support teams when issues pop up.
Would need some elaboration in that area, best if there is some specific issue already registered that would be a problem for the support team asking users for `/usr/bin/env perl -v` versus `/usr/bin/perl -v` to get a specific version - as I assume any version-based issue would require to ask user anyway to run a set of diagnostic commands to get a full picture of user env (since CLI relies on user env setup anyway).

patryk-kozak:
I am going to close that PR, I am gonna patch everything on the fork under https://github.com/patryk-kozak/genesis/tree/fix-shebangs-use-env so I can unblock myself.

PR #477, opened 2022-07-18
Enhance the PIPELINES.md readme

patryk-kozak:
Intention behind this change is to enhance a readme for PIPELINE.md to include navigational ToC and to add some of the commands that are not necessary needed for `genesis repipe` but are crucial to configure pipeline first.

### DIFF
--- a/docs/PIPELINES.md
+++ b/docs/PIPELINES.md
@@ -1,6 +1,26 @@
 Pipelines
 =========
 
+- [Pipelines](#pipelines)
+  * [Propagation](#propagation)
+  * [Configuration](#configuration)
+    + [Schema](#schema)
+      - [Pipeline YAML Field Guide](#pipeline-yaml-field-guide)
+      - [Secrets](#secrets)
+      - [Git Integration](#git-integration)
+      - [Locker Integration](#locker-integration)
+      - [Auto-Update Genesis Assets](#auto-update-genesis-assets)
+      - [Notifications](#notifications)
+        * [Notification: Slack](#notification--slack)
+        * [Notification: Email](#notification--email)
+      - [BOSH Associations](#bosh-associations)
+      - [Task Configuration](#task-configuration)
+      - [Groups](#groups)
+      - [Pipeline Override Configuration](#pipeline-override-configuration)
+    + [Useful commands](#useful-commands)
+      - [Retrieve **bosh** configuration](#retrieve---bosh---configuration)
+      - [Retrieve **vault** configuration](#retrieve---vault---configuration)
+
 Concourse is an integral part of Genesis v2.  But rather than force operators
 to define the full structure of a Concourse deployment pipeline (which can get
 rather complex), we want to provide them a domain-specific language for
@@ -300,7 +320,7 @@ pipeline:
 #### Secrets
 
 - **pipeline.vault.url** - The URL of your Vault installation, i.e.
-  `https://vault.example.com`.  This is **required**.
+  `https://vault.example.com` or `https://10.4.1.7`.  This is **required**.
 
 - **pipeline.vault.role** - The AppRole GUID of a given Vault AppRole, used to
   generate temporary tokens for Vault-accessing during deploys. This field is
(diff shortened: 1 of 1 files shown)


docs/PIPELINES.md
```diff
@@ -520,18 +540,69 @@ being deployed, or vice versa.
 
 If you need to edit and make changes to your deployment pipeline, simply add
 the changes you need to the bottom of your ci.yml located in your
-*-deployments repo. For example, if we have a pipeline we need to edit the
+*-deployments repo to be merged by spruce. 
+For example, if we have a pipeline we need to edit the
 'check_every:' parameter of a resource named git we can add the block below
-under our pipeline layouts:
+under our pipeline:
```
dennisjbell, line 546: why was `layouts` removed?  It was referring to the layouts section.

docs/PIPELINES.md
```diff
@@ -520,18 +540,69 @@ being deployed, or vice versa.
...
-  test-sandbox-ops: |+
-    auto *test-sandbox *test-staging
-    test-sandbox -> test-staging
+  (...)
+  layouts:
+    test-sandbox-ops: |+
+      auto *test-sandbox *test-staging
+      test-sandbox -> test-staging
+
+  groups:
```
dennisjbell, line 556: While this group section is valid, it doesn't follow from the example about that talks about adding check-every after the layouts section...  perhaps it should be noted in a second example.

docs/PIPELINES.md
```diff
@@ -520,18 +540,69 @@ being deployed, or vice versa.
...
+```
+`[env]` - name of the env/yaml file
+
+You should see that env `url` `username` `password` and `ca certificate`. Best approach is to put them into the Vault vs plaintext in `ci.yml`.\
+By convention they should be stored under something similar to:
+```bash
+secret/[env]/bosh/ssl/ca:certificate
+secret/[env]/bosh/users/[user]:password
+```
+Make sure crendtials are there before applying!
```
dennisjbell, line 587: typo: crendtials

docs/PIPELINES.md
```diff
@@ -520,18 +540,69 @@ being deployed, or vice versa.
...
+    - test-staging
 
 resources:
 - name: git
   check_every: 15m
 ```
+
+### Useful commands
+
+#### Retrieve **bosh** configuration
```
dennisjbell, line 568: This may no longer be needed, as genesis now grabs the bosh from the exodus data.  Furthermore, the password and cert, even if needed, should come from the concourse integrated vault.  Lets confirm these two things and get updated docs.

docs/PIPELINES.md
```diff
@@ -520,18 +540,69 @@ being deployed, or vice versa.
...
+
+You should see that env `url` `username` `password` and `ca certificate`. Best approach is to put them into the Vault vs plaintext in `ci.yml`.\
+By convention they should be stored under something similar to:
+```bash
+secret/[env]/bosh/ssl/ca:certificate
+secret/[env]/bosh/users/[user]:password
+```
+Make sure crendtials are there before applying!
+
+#### Retrieve **vault** configuration
```
dennisjbell, line 589: The genesis-pipeline app role should be autodiscovered by the pipeline.  If it is not, that's a bug.  We should document how that approle gets generated and made available though.

dennisjbell (changes requested):

bgandon:
I suggest taking over this PR and suggest improvements and fixes on Patrick's work, as he has switched to other projects.

docs/PIPELINES.md
bgandon: I've reworked the text.

docs/PIPELINES.md
bgandon: I've simplified the example and replaced that with an ellipsis.

docs/PIPELINES.md
bgandon: indeed, good catch

docs/PIPELINES.md
bgandon: I've added your remark as a note, with the idea in mind that we also need to document older way to do things for older Genesis installations.

docs/PIPELINES.md
bgandon: I've also added your remark as a note, just like with the `boshes` section, with the idea in mind that we also need to document older way to do things for older Genesis installations.

bgandon:
Rebase this work on top of the `v2.8.x-dev` branch.

PR #483, opened 2022-08-09
Bosh debug ops

dennisjbell:
[Bug Fixes]

* When using legacy secrets provider, if the current safe target has the
  same URL as another safe target, a confusing error would be printed.
  This change provides a much more informative error message.

* When using a specified secrets provider for a deployment repo, a more
  instructive error message is provided when multiple safe targets with
  the same URL is specified.

[Improvements]

* `genesis bosh` now supports all the basic options, including
  --[no-]color, -q|--quiet, -D|--debug, -T|--trace, and -D|--show-stack.

### DIFF
--- a/bin/genesis
+++ b/bin/genesis
@@ -1315,11 +1315,11 @@ sub {
 
 	explain("${ui_nl}Secrets provider for #C{%s} deployment at #M{%s}:", $top->type, $top->path);
 	my %vault_info = $top->vault_status;
-	if ($vault_info{status} eq "unauthenticated") {
-		eval {$top->vault->authenticate}; # Try to auto-authenticate
-		%vault_info = $top->vault_status;
-	}
 	if (%vault_info) {
+		if ($vault_info{status} eq "unauthenticated") {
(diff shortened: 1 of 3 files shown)


wayneeseguin (commented):
Why is `-C, --cwd` removed from the help output now? Looks like it's still an option on line 1752, or was that intended to be removed also?
I don't see any issues with the rest of it.

dennisjbell:
> 



> Why is `-C, --cwd` removed from the help output now? Looks like it's still an option on line 1752, or was that intended to be removed also? I don't see any issues with the rest of it.

The $REPO_USAGE on https://github.com/genesis-community/genesis/pull/483/files#diff-aee325df821eada8817a389cf7f137562564f10809f6e22a2b45de41f4c33647R1730 contains the usage output for --help, -C|--cwd and the other added options

wayneeseguin:
> 

Clever - OK then LGTM!

wayneeseguin (approved):
Comments resolved LGTM

PR #487, opened 2022-08-21
Unset https_proxy env var for bosh calls

### LINKED ISSUE #484: proxy must be :ENV or a URI::HTTP
When using an Ubuntu Jammy Bastion, the initial deploy of proto-bosh fails on Stemcell upload with the following error message:

```
creating stemcell (bosh-azure-hyperv-ubuntu-bionic-go_agent 1.61):
  CPI 'create_stemcell' method responded with error: CmdError{"type":"Bosh::Clouds::CloudError","message":"Failed to upload page bl``create_page_blob: inspect:#\u003c
	ArgumentError: proxy must be :ENV or a URI::HTTP\u003e
	backtrace:/home/ops/.bosh/installations/383443bb-01a1-4a01-4795-bd614b402691/packages/bosh_azure_cpi/vendor/bundle/ruby/3.1.0/gems/net-http-persistent-4.0.1/lib/net/http/persistent.rb:756:in `proxy='
/home/ops/.bosh/installations/383443bb-01a1-4a01-4795-bd614b402691/packages/bosh_azure_cpi/vendor/bundle/ruby/3.1.0/gems/faraday-net_http_persistent-1.2.0/lib/faraday/adapter/net_http_persistent.rb:26:in `net_http_connection'
```

Genesis CLI sets `http_proxy` and `https_proxy` to `""` when it has not been configured in the manifest which ultimately trickles down into how this version of Net::HTTP::Persistent handles the incoming proxy which fails to exclude "blank" configurations:

```
    @proxy_config = case proxy
       when :ENV      then Net::HTTP::Configuration.new(:proxy => :environment)
       when URI::HTTP then Net::HTTP::Configuration.new(:proxy => proxy.to_s)
       when nil       then # ignore
       else raise ArgumentError, 'proxy must be :ENV or a URI::HTTP'
    end
```

The error condition prevents the stemcell upload from succeeding.

#### Workaround
Deploying a simple squid proxy on the localhost and then setting `http_proxy` and `https_proxy` to pass traffic through allows the upload to succeed.

#### Suggested Fix
Do not set the environment variable for `http_proxy` and `https_proxy` when they are not set in the manifest.

dennisjbell:
[Bug Fixes]

* Genesis doesn't like the `https_proxy` or `HTTPS_PROXY` environment
  variables being set, but BOSH was happy when they were set to "".
  However, upstream BOSH now considers that an error, so Genesis is
  going to simply unset them.

  If you need to use HTTPS proxy, you can still set `GENESIS_HONOR_ENV`
  to a truthy value, and it will keep `HTTPS_PROXY` and `https_proxy`
  environment variables from the calling scope, but only do this if
  you encounter an error.

---

Fixes #484

### DIFF
--- a/lib/Genesis/BOSH.pm
+++ b/lib/Genesis/BOSH.pm
@@ -86,8 +86,8 @@ sub execute {
 	$env{$_} = $opts->{env}{$_} for (keys %{$opts->{env}||{}});
 	$opts->{env} = \%env;
 	unless ($ENV{GENESIS_HONOR_ENV}) {
-		$opts->{env}{HTTPS_PROXY} = ''; # bosh dislikes this env var
-		$opts->{env}{https_proxy} = ''; # bosh dislikes this env var
+		$opts->{env}{HTTPS_PROXY} = undef # bosh dislikes this env var
+		$opts->{env}{https_proxy} = undef # bosh dislikes this env var
 	}
 	my $noninteractive = envset('BOSH_NON_INTERACTIVE') ? ' -n' : '';
 


tpoland (approved):
LGTM

PR #488, opened 2022-08-22
Added alternative vault for specified bosh env

dennisjbell:
[Improvements]

* You can now specify where your alternative BOSH director's credentials
  can be found.

  By convention, your BOSH environment (the director you are using to
  deploy your environment) is the same name as the environment you are
  deploying, and its credentials are stored in `exodus` in the same
  vault.

  This isn't possible when what you're deploying is another BOSH
  director, so Genesis has always made it possible to specify another
  BOSH environment alias, via the `genesis.bosh_env` parameter, and
  before that, via the `params.bosh` in the bosh kit.

  Furthermore, if you deployed your BOSH director with a non-standard
  kit, you could specify that deployment-type as such:
  `<alias>/<deployment-type>`.  By default, that deployment type is
  `bosh`

  When v2.8.0 came around, we started storing access for the credentials
  to the BOSH director in vault.  While we always recommend using a
  single vault for all deployments in your deploy repo, some companies
  have policies that don't allow that.  To this end, you can now specify
  which vault and exodus base path under which your BOSH director
  credentials can be found.  This extends `genesis.bosh_env` with the
  following pattern:

  ```
  <alias>[/<deployment-type>][@[http(s?)://<host>[:<port]/]<base-path>]
  ```

  We still recommend a single vault and single exodus base path whenever
  possible, as most of the existing kits that use exodus data do not
  have the ability to specify an alternative vault or path.  This simply
  allows the layering of nested BOSHes to work within the confines of
  corporate security policies.

### DIFF
--- a/lib/Genesis/Env.pm
+++ b/lib/Genesis/Env.pm
@@ -1131,11 +1131,23 @@ sub bosh {
 			}
 		}
 
-		my ($bosh_alias,$bosh_dep_type) = split('/', $self->bosh_env);
+		# bosh env can be <alias>[/<deployment-type>]@[http(s?)://<host>[:<port>]/][<mount>]
+		my ($bosh_alias,$bosh_dep_type,$bosh_exodus_vault,$bosh_exodus_mount) =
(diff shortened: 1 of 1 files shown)


dennisjbell:
I can as soon as the reviews are approved.

wayneeseguin (approved):

PR #491, opened 2022-08-28
Added `genesis manifest --bosh-vars`

dennisjbell:
Not for general use, though nothing is stopping it being used by end
users, this feature is intended to provide the contents of bosh
variables requirement for the forthcoming `genesis remove`


tpoland (approved):
LGTM

krutten (approved):

PR #494, opened 2022-09-21
Working implementation for email notifications

bgandon:
Hi,

In this PR I've reworked a bit the way email notifications are implemented in the Genesis-generated pipelines, when some `pipeline.email` hash is specified in the `ci.yml` of some deployment.

You'll notice that I've got rid of the `build-email-*` script resources. I must say that modeling a script generation using a resource just doesn't fit with what Concourse mean by "resource"—It's supposed to be something that can have a new version (`check`) and from which one can get that version (`get`) or create a new one (`put`). Just using some ad-hoc Bash code just looks better.

The `starkandwayne/concourse` image is hardwired and possibly should rely on `$pipeline->{pipeline}{task}{image}` like elsewhere in the code. Don't hesitate to mention that in your review.

Also the Bash script code has ugly escape sequences. This come from the fact that the `_gen_notifications` subroutine is to be agnostic of the YAML indentation level from where it is invoked in the YAML tree. If you insist, I can try to write this script as a multi-line YAML string and see if it works. That would read a little bit better, I admit.

About the `X-Concourse-Site-Env:` MIME header, I didn't get it why it was here. I've also wondered where the `${CI_SITE_ENV}` environment variable was supposed to come from, without being able to locate anything in Genesis or Concourse. So, I've dropped the feature here, that I've guessed to be experimental. We can always implement it again properly when we know more about it.

Best,
Benjamin

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,3 +18,20 @@ your interactions with the project.
 
 3. After review and approval, your Pull Request will be merged by
    a repository owner.
+
+## Test your changes
+
+In order to test your changes, you can use the following development workflow:
+
+- Clone this repo
+- Do some changes
+- Build a development `genesis` CLI running `make release VERSION=x.y.z`
+- Symlink the generated `genesis-x.y.z` (or `genesis-x.y.z-dirty` if you
+  didn't commit your code yet) file to `genesis` with
+  `ln -s genesis-x.y.z-dirty genesis` (to be done once only)
+- Add the current directory in your `PATH` with `export PATH=$PWD:$PATH` (to
+  be done only once per shell session) or just copy it to you `~/bin`
+  directory if it exists and is on your path
+- Go to some deployment directory, check the genesis CLI you'll be using with
+  `which genesis` and `genesis version`
+- Run the usual `genesis` CLI commands and verify it behaves as expected
--- a/lib/Genesis/CI/Legacy.pm
+++ b/lib/Genesis/CI/Legacy.pm
@@ -63,37 +63,33 @@ EOF
 		$notification .= <<EOF;
 {
   do: [
-  { get: build-email-$alias },
   { task: write-email-body,
+    params: {
+      PIPELINE_NAME: (( grab pipeline.name ))
+    },
     config: {
       platform: linux,
       image_resource: {
         type: registry-image,
-        source: {
-          repository: ubuntu,
-        },
+        source: { repository: starkandwayne/concourse },
       },
-
-      inputs: [
-        { name: build-email-$alias },
-        { name: out },
-      ],
       outputs: [
         { name: email },
       ],
-
       run: {
-        path: build-email-$alias/run,
-        args: [],
+        path: bash,
+        args: [
+          "-exc",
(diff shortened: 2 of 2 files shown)


lib/Genesis/CI/Legacy.pm
```diff
@@ -1099,23 +1093,6 @@ EOF
...
-  - name: build-email-changes-staged
-    type: script
-    icon: email-outline
-    source:
-      filename: run
-      body: |
-        #!/bin/bash
-        mkdir -p email
-        rm -rf email/*
-        echo "X-Concourse-Site-Env: ${CI_SITE_ENV}" >>email/header
```
wayneeseguin, line 1111: Does the new method without this still convey the context of where the notification is coming from? This has the `${CI_SITE_ENV}` as part of that.

bgandon: All right, I can have a look and propose a solution for this concern.

bgandon: We have the pipeline name in the subject of the e-mail notification:
https://github.com/gstackio/genesis/blob/eb1ec5aafd/lib/Genesis/CI/Legacy.pm#L83

And we have the environment names in the text body of the email:
https://github.com/gstackio/genesis/blob/eb1ec5aafd/lib/Genesis/CI/Legacy.pm#L1225-L1236

That's a bit different from other instant message notifications where we prepend the same message with `<PIPELINE-NAME>: `, but results in providing similar information.

Do you think we are missing any relevant information here, like the URL of Concourse or anything? Just let me know.

wayneeseguin: If it's easy to add the url to the pipeline so ops can click on it and go to the context that's great. If not this is good enough, LGTM.

dennisjbell: The code already covers the context of the build that passed/failed in the contents of $message, so we should be good, though I can see the desire to have the subject contain at least the job that failed for easy reference without opening the mail.  That could be a future nice-to-have.

CONTRIBUTING.md
```diff
@@ -18,3 +18,18 @@ your interactions with the project.
...
+## Test your changes
+
+In order to test your changes, you can use the following development workflow:
+
+- Clone this repo
+- Do some changes
+- Build a development `genesis` CLI running `make release VERSION=x.y.z`
+- Symlink the generated `genesis-x.y.z-dirty` file to `genesis` with
+  `ln -s genesis-x.y.z-dirty genesis` (to be done once only)
+- Add the current directory in your `PATH` with `export PATH=$PWD:$PATH` (to
```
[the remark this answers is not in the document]
bgandon: I'll mention this too 👌

lib/Genesis/CI/Legacy.pm
bgandon: Yes, the Concourse URL is transmitted to the `genesis repipe` invocation through the `-t` target, so it's to be grabbed from `~/.flyrc` targets. Does this fall into what you would qualify as “easy”, Wayne? Or put in another way, would we allow us to grab that URL from there?

As other types of notifications like Slack are also missing this URL, I would argue for for a future evolution. Same for failing job in the subject.

bgandon:
I've answered your remarks and pushed some adjustments to the docs. I let you guys have a look and approve if it's OK to you.

xiujiao (approved):

PR #497, opened 2022-10-14
Suggest some documentation for implicit features

bgandon:
Hi there,

After discussing with Wayne about _virtual features_, I've suggested an alternate naming of _implicit features_ that he has liked very much.

Then I figured out that I couldn't find any documentation about those, so here in this PR I'm suggesting some.

Don't hesitate to point out any other place where some previous docs would have to be updated.

Best,
Benjamin


norman-abramovitz (approved):
LGTM

PR #498, opened 2022-10-28
Support for private registry in generated pipelines, for air-gapped environments

bgandon:
Hi,

In this PR, I've introduced 3 new configs `pipeline.registry.{uri,username,password}` so that pipelines in air-gapped environments can fetch their custom resource types from a private registry.

For task images, I've chosen to forcefully prepend the registry prefix to the image repo whenever the registry is defined, assuming that operators wouldn't need to fetch images from different private registries.

Don't hesitate to comment.

Best,
Benjamin


norman-abramovitz (approved):
LGTM

PR #501, opened 2022-11-10
Escape registry password

bgandon:
This brings some improvement on top of the #498 new feature.

In case it may contain characters that conflict with the YAML syntax, which has been the case with the pipelines we are currently generating.


bgandon:
We've tested that code in the customer environment, pushed some fixes and it works fine.

norman-abramovitz (approved):
LGTM

dennisjbell (approved):

dennisjbell:
Acccidentally merged into ocfp-pipeline after retrofitting everything back into v2.8.x-dev -- manually cherrypicked 4e8af54 into v2.8.x-dev

PR #505, opened 2022-11-16
Add missing Vault env vars for running 'genesis ci-pipeline-run-errand'

bgandon:
Hi there,

Seems the Vault credentials have been missing when running errand from generated pipelines.

Please note that this code is not tested yet. I'll update here when it is.

Best,
Benjamin


bgandon:
I thought about integrating this patch in another PR but finally it's better submitted standalone.

dennisjbell:
Seems to already be merged

PR #506, opened 2022-11-17
Improve YAML-encoding in generated pipelines

bgandon:
Hi there,

This PR improves the way registry and Git passwords are YAML-encoded when generating pipeline.

This time, the ad-hoc implementation is replaced by the solid `encode_json()` function from the `JSON::PP` module, that is already used elsewhere in Genesis.

Best,
Benjamin


dennisjbell (approved):
lgtm

PR #507, opened 2022-11-17
Proper secrets mount in generated pipelines

bgandon:
Hi there,

Seems the Vault credentials have been missing when running errand from generated pipelines. We haven't tested it unfortunately, as we still don't have the CF pipeline working, which should happen soon. In the meantime @dennisjbell could you confirm that these Vault environment variables are indeed required?

Then, we've added the `GENESIS_SECRETS_MOUNT` env var, so that `genesis` invocations in pipeline task get their `genesis.secrets_mount` forced to the correct value.

Please note that this code is not tested yet, thus the Draft PR. I'll update here when it is.

Best,
Benjamin


bgandon:
Canceling this PR in favor of #508.

PR #508, opened 2022-11-23
Vault fixes

bgandon:
This PR contains the fixes that we've made when pairing together the passed days.

Plus some more I've spotted that required similar fixes.


dennisjbell (approved):
lgtm

